PR #2, opened 2023-11-03
Feat/sms batches: `Send` & `DryRun`

JPPortier:
- Implement SMS batches `Send` and `DryRun`
- Merge/PR for `sms` branches and not main: `sms` feature branch will be merge to main only when full set of SMS features will be available

### DIFF
--- a/client/src/main/com/sinch/sdk/domains/sms/models/DryRunPerRecipientDetails.java
+++ b/client/src/main/com/sinch/sdk/domains/sms/models/DryRunPerRecipientDetails.java
@@ -1,7 +1,7 @@
 package com.sinch.sdk.domains.sms.models;
 
 /**
- * Dtails related to a dryRun recipient
+ * Details related to a dryRun recipient
  *
  * @since 1.0
  */
--- a/client/src/main/com/sinch/sdk/domains/sms/models/DryRunPerRecipientDetails.java
+++ b/client/src/main/com/sinch/sdk/domains/sms/models/DryRunPerRecipientDetails.java
@@ -1,0 +1,94 @@
+package com.sinch.sdk.domains.sms.models;
+
+/**
+ * Dtails related to a dryRun recipient
+ *
+ * @since 1.0
+ */
+public class DryRunPerRecipientDetails {
+
+  private final String recipient;
+  private final String messagePart;
+  private final String body;
+  private final String encoding;
+
+  public DryRunPerRecipientDetails(
+      String recipient, String messagePart, String body, String encoding) {
+    this.recipient = recipient;
+    this.messagePart = messagePart;
+    this.body = body;
+    this.encoding = encoding;
+  }
+
+  public String getRecipient() {
+    return recipient;
+  }
+
+  public String getMessagePart() {
+    return messagePart;
+  }
+
+  public String getBody() {
+    return body;
+  }
+
+  public String getEncoding() {
+    return encoding;
+  }
+
+  @Override
+  public String toString() {
+    return "DryRunPerRecipientDetails{"
+        + "recipient='"
+        + recipient
+        + '\''
+        + ", messagePart='"
+        + messagePart
+        + '\''
+        + ", body='"
+        + body
+        + '\''
+        + ", encoding='"
+        + encoding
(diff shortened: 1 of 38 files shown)


client/src/main/com/sinch/sdk/domains/sms/models/BatchText.java
```diff
@@ -9,12 +9,12 @@
...
  */
 public class BatchText extends Batch<String> {
-  private final boolean flashMessage;
+  private final Boolean flashMessage;
   private final Parameters parameters;
-  private final boolean truncateConcat;
-  private final int maxNumberOfMessageParts;
-  private final int fromTon;
-  private final int fromNpi;
+  private final Boolean truncateConcat;
```
asein-sinch, line 14: Why do you need to use Objects instead of primitives?

client/src/main/com/sinch/sdk/domains/sms/models/DryRunPerRecipientDetails.java
```diff
@@ -0,0 +1,94 @@
+package com.sinch.sdk.domains.sms.models;
+
+/**
+ * Dtails related to a dryRun recipient
```
asein-sinch: Typo 🙃

core/src/main/com/sinch/sdk/core/exceptions/ApiExceptionBuilder.java
```diff
@@ -54,4 +54,26 @@ private static Optional<ApiException> getExceptionFromNumbersOrSmsError(
...
+    //     "text": string,
+    // }
+
+    if (null == mappedResponse) {
+      return Optional.empty();
+    }
+
+    Object code = mappedResponse.get("code");
+    Object text = mappedResponse.get("text");
+    ;
```
asein-sinch: A trailing semicolon to remove

asein-sinch (approved):
It's nice to see generated files updated after you changed the mustache template!

client/src/main/com/sinch/sdk/domains/sms/models/BatchText.java
JPPortier: Due to autoboxing.
When receiving a DTO instance where fields is not present or null: autoboxing is causing NullPointerException to map missing value to primitive.
e.g.: Sending a SMS with only `to`and `body` field (see sample-app `Send` source)
REST API is returning:
```
{
    "id": "01HEAC0AG69SVYYQ675VPYT28Q",
    "to": [
        "XXXXX"
    ],
    "canceled": false,
    "body": "the body",
    "type": "mt_text",
    "created_at": "2023-11-03T10:35:03.558Z",
    "modified_at": "2023-11-03T10:35:03.558Z",
    "delivery_report": "none",
    "expire_at": "2023-11-06T10:35:03.558Z",
    "feedback_enabled": false,
    "flash_message": false
}
```
Then BatchText.truncatConcat as boolean primitive cannot contains "valid" information (nor true, nor false) and autoboxing during convertion from DTO(https://github.com/sinch/sinch-sdk-java/pull/2/files#diff-45ba5a02d31c9bfab5d5e2a1c03fdf3809bc78697f9dcbbffbc9c0bc6652f4a2R110) is throwing the exception because:
- `dto.getTruncateConcat()` is returning `null` (field is a `Boolean` and not `boolean`)
- then `null` cannot be converted to a boolean because of null value
- then NullPointerException

Instead of modifying DTO object to use primitives, I'm using Object to be able have distinction between what was recevied/set and what was not.
In addition if there is some default value to be set; then it is to the server side to set this default value on return.

core/src/main/com/sinch/sdk/core/exceptions/ApiExceptionBuilder.java
JPPortier: Fixed

client/src/main/com/sinch/sdk/domains/sms/models/DryRunPerRecipientDetails.java
JPPortier: Fixed

PR #5, opened 2023-11-07
SMS: webhooks

JPPortier:
Because of java is not having a JSON parser by default: we provide a `webHooks` service to parse them and return dedicate object instances.
This helper avoid to SDK users to:
- instantiate and use themselves a jackson parser
- have direct Jackson dependency 
- let SDK implementation to handle de-serialization process

### DIFF
--- a/client/src/main/com/sinch/sdk/domains/sms/adapters/WebHooksService.java
+++ b/client/src/main/com/sinch/sdk/domains/sms/adapters/WebHooksService.java
@@ -23,6 +23,7 @@ public class WebHooksService implements com.sinch.sdk.domains.sms.WebHooksServic
     }
   }
 
+  @Override
   public BaseDeliveryReport deliveryReport(String jsonPayload) throws ApiMappingException {
     try {
       BaseDeliveryReport generic =
--- a/client/src/main/com/sinch/sdk/domains/sms/adapters/WebHooksService.java
+++ b/client/src/main/com/sinch/sdk/domains/sms/adapters/WebHooksService.java
@@ -1,0 +1,38 @@
+package com.sinch.sdk.domains.sms.adapters;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.sinch.sdk.core.exceptions.ApiMappingException;
+import com.sinch.sdk.core.utils.StringUtil;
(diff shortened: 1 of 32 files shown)


client/src/main/com/sinch/sdk/domains/sms/adapters/WebHooksService.java
```diff
@@ -0,0 +1,38 @@
...
+      if (null == generic && !StringUtil.isEmpty(jsonPayload)) {
+        throw new ApiMappingException(jsonPayload, null);
+      }
+      return generic;
+    } catch (JsonProcessingException e) {
+      throw new ApiMappingException(jsonPayload, e);
+    }
+  }
+
+  public BaseDeliveryReport deliveryReport(String jsonPayload) throws ApiMappingException {
```
asein-sinch, line 27: This method can have the `@Override` tag too

sample-app/src/main/java/com/sinch/sample/sms/webhooks/DeliveryReport.java
```diff
@@ -0,0 +1,45 @@
...
+            + "        \"44231235674\"\n"
+            + "      ],\n"
+            + "      \"status\": \"Delivered\"\n"
+            + "    }\n"
+            + "  ],\n"
+            + "  \"total_message_count\": 1\n"
+            + "}";
+    LOGGER.info("Convert payload" + payload);
+
+    Object value = client.sms().webHooks().deliveryReport(payload);
```
asein-sinch: Is it possible to get a class other than `Object`? The interface mentions `BaseDeliveryReport` but I'm wondering if the end user should be exposed to the `Base` class naming

JPPortier: You're right,
Fixed.
Note: the end user will have to use `instanceOf` onto result (if required for further processing) to detect the proper type if parsed payload (this is intentional)

client/src/main/com/sinch/sdk/domains/sms/adapters/WebHooksService.java
JPPortier: Fixed

650elx (approved):
Looks clean :)

PR #6, opened 2023-11-09
Sms delivery reports

JPPortier:
- moved WebHook classes for Delivery Report at "SMS level" because of shared with Delivery Report endpoints and not only used by hooks
- Refactored `DeliveryReport` enum for `type` to `DeliveryReportType`
- consume generated DTO file for SMS WebHooks service
- Implement `get` / `getForNumber` and `list`


client/src/main/com/sinch/sdk/domains/sms/adapters/DeliveryReportsService.java
```diff
@@ -0,0 +1,109 @@
...
+ * a delivery report</a> or sent as a callback.
+ *
+ * @see <a
+ *     href="https://developers.sinch.com/docs/sms/api-reference/sms/tag/Delivery-reports">https://developers.sinch.com/docs/sms/api-reference/sms/tag/Delivery-reports</a>
+ * @since 1.0
+ */
+public class DeliveryReportsService implements com.sinch.sdk.domains.sms.DeliveryReportsService {
+
+  private Configuration configuration;
+  private DeliveryReportsApi api;
```
Dovchik, line 39: `final` allows to set the variable only once anywhere, be it constructor, or another method. Does it makes sense to mark this variables as `final` in this context?

Dovchik (approved):

client/src/main/com/sinch/sdk/domains/sms/adapters/DeliveryReportsService.java
JPPortier: It should yes. 
But I'm facing an issue (not yet finalized) due to unit testing and mocking library: when set to `final` mocking lib is not able to set variables and test execution are facing null pointer.
I'll have to take time to fix this...

PR #13, opened 2023-11-21
AUTH manager refactoring

JPPortier:
Dissociate the OAS alias names for security schemes from the authorized scheme keyword  (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml)

OAS files are having security table:
```
security:
  - BearerAuth: []
  - 
```
to be mapped to a `Security schemes` section
```
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
```
So `BearerAuth` is a free form string (from security) and `Bearer` have to be in list from https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

Refactoring code  prepare next free form name coming from 'Verification': `Basic`(currently Numbers is using `BasicAuth`) and will authorized single AuthManager with different alias from OAS files

### DIFF
--- a/client/src/main/com/sinch/sdk/SinchClient.java
+++ b/client/src/main/com/sinch/sdk/SinchClient.java
@@ -36,6 +36,12 @@ public class SinchClient {
 
   private HttpClientApache httpClient;
 
+  // TODO: generator tools do not generate security schemas related information.
+  // Will have to fix it with a generic way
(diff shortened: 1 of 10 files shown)


JPPortier:
Closing PR: will move to another pattern due to Verification API "Basic" support won't fit with current unified authentication mechanism
This refactoring do not have any utility in this context

PR #14, opened 2023-11-21
Feat: auth manager refactoring enabling to have dedicated managers to each underline APIs

JPPortier:
This mean: auth managers no longer attached at HttpClient level, this to avoid conflicts naming from cross OAS file aliases.
Keep a single HttpClient instance shared across all APIs handled by SDK


client/src/main/com/sinch/sdk/http/HttpClientApache.java
```diff
@@ -123,14 +138,35 @@ public HttpResponse invokeAPI(ServerConfiguration serverConfiguration, HttpReque
...
-  private boolean processUnauthorizedResponse(HttpRequest request, HttpResponse response) {
+  private boolean processUnauthorizedResponse(
+      HttpRequest request,
+      HttpResponse response,
+      Map<String, AuthManager> authManagersByOasSecuritySchemes) {
+
+    Map<String, AuthManager> authManagersByAuthSchemes =
+        authManagersByOasSecuritySchemes.values().stream()
+            .map(authManager -> new AbstractMap.SimpleEntry<>(authManager.getSchema(), authManager))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (a1, a2) -> a1));
```
asein-sinch, line 149: I think it's not needed to build this `Map` as you are just interested into the existence of an occurence of the key `SCHEMA_KEYWORD_BEARER`

asein-sinch: Arf, forget about it, the diff made me read incorrectly.

asein-sinch (approved):
LGTM ⭐️

PR #16, opened 2023-11-24
Feat/verification get report

JPPortier:


client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
```diff
@@ -71,7 +83,7 @@ private String getSignature(
         "\n",
         method,
         bodyMD5Hash,
-        httpContentType,
+        null != httpContentType ? httpContentType : "",
```
asein-sinch, line 86: Do we have cases where there is no content type defined for a request?

client/src/main/com/sinch/sdk/domains/verification/StatusService.java
```diff
@@ -0,0 +1,49 @@
...
+ * Get the status of specific verification requests in the verification process. Returns the status
+ * of pending and completed verifications. You can retrieve the status of verification requests by
+ * using the ID of the request, the phone number of the user being verified, or a custom reference
+ * string.
+ *
+ * @see <a
+ *     href="https://developers.sinch.com/docs/verification/api-reference/verification/tag/Verification-status">https://developers.sinch.com/docs/verification/api-reference/verification/tag/Verification-status</a>
+ * @since 1.0
+ */
+public interface StatusService {
```
asein-sinch, line 19: The same way you are using the same name for `report` because the signatures are different, can we imagine having just one `get` (and define classes (or records) for `Id` and `Reference`)

client/src/main/com/sinch/sdk/domains/verification/models/requests/VerificationReportFlashCallRequestParameters.java
```diff
@@ -0,0 +1,56 @@
+package com.sinch.sdk.domains.verification.models.requests;
+
+import com.sinch.sdk.domains.verification.models.VerificationMethodType;
+
+public class VerificationReportFlashCallRequestParameters
+    extends VerificationReportRequestParameters {
+
+  private final String cli;
+
+  public VerificationReportFlashCallRequestParameters(String cli) {
```
asein-sinch, line 10: The famous `cli` that we will never know about 🙃

client/src/main/com/sinch/sdk/domains/verification/models/requests/VerificationReportSMSRequestParameters.java
```diff
@@ -0,0 +1,71 @@
...
+    this.code = code;
+    this.cli = cli;
+  }
+
+  public String getCode() {
+    return code;
+  }
+
+  public Optional<String> getCli() {
+    return Optional.ofNullable(cli);
```
asein-sinch, line 22: Hopefully this one is optional 😮‍💨

openapi-contracts/src/main/com/sinch/sdk/domains/sms/adapters/api/v1/BatchesApi.java
```diff
@@ -115,10 +115,9 @@ private HttpRequest cancelBatchMessageRequestBuilder(String servicePlanId, Strin
         "/xms/v1/{service_plan_id}/batches/{batch_id}"
             .replaceAll(
                 "\\{" + "service_plan_id" + "\\}",
-                URLParameterUtils.encodeParameterValue(servicePlanId.toString()))
+                URLPathUtils.encodePathSegment(servicePlanId.toString()))
```
asein-sinch, line 118: Nice catch! Hopefully, the service plan id doesn't contain characters that would be encoded differently being a query param or a path param.

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/VerificationReportDtoTest.java
```diff
@@ -0,0 +1,90 @@
...
+          .reason("Fraud")
+          .callComplete(true)
+          .reference("my reference")
+          .price(
+              new VerificationResponsePriceDto(
+                  new VerificationPriceInformationDto()
+                      .verificationPrice(
+                          new VerificationPriceInformationVerificationPriceDto(
+                              new MoneyDto()
+                                  .currencyId("verificationPrice currency id")
```
asein-sinch, line 29: The specification saying that the currency id is a "ISO 4217 currency code", you may want to use a real one in case we implement a type validation at some point

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/VerificationReportDtoTest.java
```diff
@@ -0,0 +1,90 @@
...
+                                  .amount(3.141516F)))
+                      .terminationPrice(
+                          new VerificationPriceInformationTerminationPriceDto(
+                              new MoneyDto()
+                                  .currencyId("terminationPrice currency id")
+                                  .amount(6.626070F)))
```
asein-sinch, line 35: Another question about the price: as we have 2 different decimal prices, do you think it would be worth to provide a helper method that returns the total price (taking care about the issues related to summing floats in java)?

sample-app/src/main/java/com/sinch/sample/verification/verifications/GetReportById.java
```diff
@@ -0,0 +1,51 @@
...
+import com.sinch.sdk.domains.verification.models.VerificationMethodType;
+import com.sinch.sdk.domains.verification.models.VerificationReport;
+import com.sinch.sdk.domains.verification.models.requests.VerificationReportCalloutRequestParameters;
+import com.sinch.sdk.domains.verification.models.requests.VerificationReportFlashCallRequestParameters;
+import com.sinch.sdk.domains.verification.models.requests.VerificationReportRequestParameters;
+import com.sinch.sdk.domains.verification.models.requests.VerificationReportSMSRequestParameters;
+import java.io.IOException;
+import java.util.logging.Logger;
+
+public class GetReportById extends BaseApplication {
```
asein-sinch, line 14: I'm not sure using `Get` in the class' name is reflecting the purpose of the API as it is a 'PUT'

asein-sinch (approved):
LGTM, great work! 🌟

client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
JPPortier: Yes: for GET requests (e.g. `VerificationStatusByReference`)
So when API is calling the HTTP client (apache for this implementation)
- array list of content type is empty (https://github.com/sinch/sinch-sdk-java/blob/8cdfba59098ef119c85fda207e5a47c5208a462c/openapi-contracts/src/main/com/sinch/sdk/domains/verification/adapters/api/v1/QueryVerificationsApi.java#L105)
- then Content-type header is not present (https://github.com/sinch/sinch-sdk-java/blob/8cdfba59098ef119c85fda207e5a47c5208a462c/client/src/main/com/sinch/sdk/http/HttpClientApache.java#L224)
- then the Auth Manager handle this case

client/src/main/com/sinch/sdk/domains/verification/StatusService.java
JPPortier: I'd like too; but it won't fit.

- Record usage: not for currently java 8 to be supported 
- Overriding: String for `id` vs `reference`: it will imply for casting onto calls with possibly end user "errors" when having to call:
  - status.get((myidclass casting) "a string id")
  - status.get((myreferenceclass casting) "a string  reference")

client/src/main/com/sinch/sdk/domains/verification/models/requests/VerificationReportFlashCallRequestParameters.java
JPPortier: Right! 🤔

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/VerificationReportDtoTest.java
JPPortier: Good catch. I'll create a ticket and having to use it for Numbers API too 👍

sample-app/src/main/java/com/sinch/sample/verification/verifications/GetReportById.java
JPPortier: To be discussed;;
As end user am I aware that I'm performing a PUT HTTP request here or at high level I'm interested in to "get" a report ... ?

client/src/main/com/sinch/sdk/domains/verification/StatusService.java
JPPortier: huum.. let me check a little bit more ..
I do not want to much over-engineering around simple "String" but may be we can imagine something like fo Double, Float, ...: `static Id Id.valueOf('String" idValue)` but this will imply REST response classes have to be designed to use these types
to be continued

JPPortier: @asein-sinch please review this implementation: https://github.com/sinch/sinch-sdk-java/pull/16/commits/0a472438085a8cbe129ec049df8b4d52fabf1ada

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/VerificationReportDtoTest.java
JPPortier: For reference: https://docs.oracle.com/javase/8/docs/api/java/util/Currency.html and ticket created (DEVEXP-206)

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/VerificationReportDtoTest.java
JPPortier: We could imagine such helper; but should be common to all SDKs (having same level of service between them)
This kind of helper/feature should, may be, provided according to users future feedbacks

client/src/main/com/sinch/sdk/domains/verification/StatusService.java
asein-sinch: This looks good. I'm just wondering if it would have some value to create a super class `StringType` that would define a field `value` and would be extended by `VerificationId` and `VerificationReference` to have a common behaviour for all the wrapped strings

client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
asein-sinch: Ok, if the server is fine with having an empty value for this field, it's fine for me too ;)

client/src/main/com/sinch/sdk/domains/verification/StatusService.java
JPPortier: A question I asked myself :-)
Because of I was mimic the Float/Double/... java classes with function like `intValue`, `floatValue`... I did not retains that way and kept dedicated function names (that could have been `idValue`/`referenceValue` according to java classes naming, but stay attached to `getters` usage from SDK)

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/VerificationReportDtoTest.java
asein-sinch: Indeed, it was a wild guess, I don't know if it's a valid use case 🤷‍♂️

client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
JPPortier: It make sens not having value: 

- a `GET` request do not have body
- `Content-type` is about how to read body request content

So not required in this context

sample-app/src/main/java/com/sinch/sample/verification/verifications/GetReportById.java
asein-sinch: Well, the action here is to "report" a verification (by submitting a `code` or a `cli`) and eventually, you get a response for this action which is indeed of the same kind of response as when you send a request to fetch a status. This is why I would rather think about `ReportWithId` or `SubmitReportWithId` as a name. WDYT?

JPPortier: Yes, in this point of view, this is having sense too.
I'll change it (my preference is for `ReportWithId`)

PR #18, opened 2023-11-27
Synch with ADR pagination

JPPortier:
Create dedicated SMS cursor based navigation to fix issue due to page_size returned value cannot be used to comoute end of navigation

### DIFF
--- a/client/src/test/java/com/sinch/sdk/domains/sms/adapters/BatchesServiceTest.java
+++ b/client/src/test/java/com/sinch/sdk/domains/sms/adapters/BatchesServiceTest.java
@@ -379,7 +379,7 @@ public class BatchesServiceTest extends BaseTest {
 
     BatchesListResponse response = service.list(null);
 
-    Iterator<Batch<?>> iterator = response.autoPageIter();
+    Iterator<Batch<?>> iterator = response.iterator();
     Batch<?> batch = iterator.next();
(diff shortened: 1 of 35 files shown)


client/src/test/java/com/sinch/sdk/domains/sms/adapters/BatchesServiceTest.java
```diff
@@ -376,10 +379,18 @@ void list() throws ApiException {
...
+            eq(null),
+            eq(null),
+            eq(null),
+            eq(null),
+            eq(null)))
+        .thenReturn(listBatchesResponseDtoPage2);
     BatchesListResponse response = service.list(null);
 
-    Iterator<Batch<?>> iterator = response.autoPageIter();
+    Iterator<Batch<?>> iterator = response.iterator();
```
Dovchik, line 393: BTW worth checking if there is no additional items in the end of the iterator

Dovchik (approved):
Minor addition about checking there is no items left in iterator

client/src/test/java/com/sinch/sdk/domains/sms/adapters/BatchesServiceTest.java
JPPortier: Yes, this what is done at line 433: `Assertions.assertThat(iterator.hasNext()).isEqualTo(false);`

PR #19, opened 2023-11-27
Support application id/secret for Verification API authentication

JPPortier:
Note the `useSecrets` function defined at Verification service level;
The sample application is calling it only if dedicated Verification config is set.
Then, the sinchclient Verification calls are using a single client instance able to use Numbers/Sms or Verification without any changes.
So when Verification API will support unified credentials: just removing the call to `useSecrets` (so removing dedicated config) and Sinch credentials will be used.
No mode side effects

### DIFF
--- a/client/src/main/com/sinch/sdk/domains/verification/VerificationService.java
+++ b/client/src/main/com/sinch/sdk/domains/verification/VerificationService.java
@@ -11,14 +11,14 @@ public interface VerificationService {
 
   /**
    * Use application secret in place of unified configuration for authentication (see Sinch
-   * dashboard for details)
+   * dashboard for details) These credentials are related to Verification Apps
    *
-   * @param key see <a href="https://dashboard.sinch.com/voice/apps">dashboard</a>
-   * @param secret see <a href="https://dashboard.sinch.com/voice/apps">dashboard</a>
+   * @param key see <a href="https://dashboard.sinch.com/verification/apps">dashboard</a>
+   * @param secret see <a href="https://dashboard.sinch.com/verification/apps">dashboard</a>
    * @return service instance for project
    * @since 1.0
    */
-  VerificationService useSecrets(String key, String secret);
+  VerificationService setApplicationCredentials(String key, String secret);
 
   /**
    * Verifications Service instance
--- a/client/src/main/com/sinch/sdk/domains/verification/VerificationService.java
+++ b/client/src/main/com/sinch/sdk/domains/verification/VerificationService.java
@@ -10,6 +10,17 @@ package com.sinch.sdk.domains.verification;
 public interface VerificationService {
 
   /**
+   * Use application secret in place of unified configuration for authentication (see Sinch
+   * dashboard for details)
+   *
+   * @param key see <a href="https://dashboard.sinch.com/voice/apps">dashboard</a>
+   * @param secret see <a href="https://dashboard.sinch.com/voice/apps">dashboard</a>
+   * @return service instance for project
+   * @since 1.0
+   */
+  VerificationService useSecrets(String key, String secret);
+
+  /**
    * Verifications Service instance
    *
    * @return service instance for project
--- a/client/src/main/com/sinch/sdk/domains/verification/adapters/StatusService.java
+++ b/client/src/main/com/sinch/sdk/domains/verification/adapters/StatusService.java
@@ -14,22 +14,29 @@ import com.sinch.sdk.domains.verification.models.VerificationReference;
 import com.sinch.sdk.domains.verification.models.VerificationReport;
 import com.sinch.sdk.models.Configuration;
 import java.util.Map;
+import java.util.function.Supplier;
 
 public class StatusService implements com.sinch.sdk.domains.verification.StatusService {
 
-  private QueryVerificationsApi api;
-
-  public StatusService() {}
+  private final Configuration configuration;
+  private final HttpClient httpClient;
+  private final Supplier<Map<String, AuthManager>> authManagerSupplier;
 
   public StatusService(
-      Configuration configuration, HttpClient httpClient, Map<String, AuthManager> authManager) {
-    this.api =
-        new QueryVerificationsApi(
-            httpClient, configuration.getVerificationServer(), authManager, new HttpMapper());
+      Configuration configuration,
+      HttpClient httpClient,
+      Supplier<Map<String, AuthManager>> authManagerSupplier) {
+    this.configuration = configuration;
+    this.httpClient = httpClient;
+    this.authManagerSupplier = authManagerSupplier;
   }
 
-  private QueryVerificationsApi getApi() {
-    return this.api;
+  protected QueryVerificationsApi getApi() {
+    return new QueryVerificationsApi(
+        httpClient,
+        configuration.getVerificationServer(),
+        authManagerSupplier.get(),
+        new HttpMapper());
   }
 
   public VerificationReport get(Identity identity, VerificationMethodType method) {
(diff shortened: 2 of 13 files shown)


client/src/main/com/sinch/sdk/domains/verification/VerificationService.java
```diff
@@ -9,6 +9,17 @@
...
+  /**
+   * Use application secret in place of unified configuration for authentication (see Sinch
+   * dashboard for details)
+   *
+   * @param key see <a href="https://dashboard.sinch.com/voice/apps">dashboard</a>
+   * @param secret see <a href="https://dashboard.sinch.com/voice/apps">dashboard</a>
+   * @return service instance for project
+   * @since 1.0
+   */
+  VerificationService useSecrets(String key, String secret);
```
asein-sinch: Defining this method at `VerificationService` level means it will have to be defined again in the `VoiceService` and the user will have to enter again the application secrets.
This is strong mapping with the split in several OAS documents and domains: from an infrastructure point of view, the applications (holding the credentials) are simply belonging to an organisation, and can be accessed by any domain that needs them.

Dovchik: What happens when user calls `verification()` without first calling this method?

JPPortier: Currently, the server will return a 401 / Authentication error because of the unified credentials will be used ... but not recognized by server

JPPortier: Exact.
Sinch dashboard is displaying twice this information (once from `Verification` tab, and once, again, onto `Voice` tab) and this is reflected here... even if information are shared.
So, if moved at an upper level (SinchClient configuration) this mean we will have
- unified credentials
- SMS credentials
- "something embedding Verification & Voice & ...)" credentials

So to rephrase: all credentials at same level: the main configuration

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationService.java
```diff
@@ -22,28 +24,33 @@ public class VerificationService implements com.sinch.sdk.domains.verification.V
...
   public VerificationService(Configuration configuration, HttpClient httpClient) {
     this.configuration = configuration;
     this.httpClient = httpClient;
 
+    // by default, unified id/secret from configuration are used but can be super sed if
+    // "useSecrets"
+    // is called after initialization
+    useSecrets(
+        configuration.getKeyId(),
+        Base64.getEncoder()
```
asein-sinch: I'm not sure we need to encode the secret as it is already provided as encoded in the applications dashboard. I don't think a customer will base64-decode it before using it in the application.

JPPortier: This code branch is for using the unified sinch credentials and in they are not base64 encoded.

Currently it do not work on runtime because server API won't accept it we need to call the dedicated useSecrets function to set proper application credentials; but if they was accepted by server, it could be the expected workflow

JPPortier:
At this time I think we are not able to agree how to "fix" the not-so-unified-way to use credentials for APIS.
dotnet is having it own implementation and a dedicated ADR is currently still in review.

So my proposal is accept the current implementation as a possible way for implementation.
This will enable to release the "Verification" feature, like the dotnet with it own implementation; and going back to refactor when ADR will be approved.

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationService.java
```diff
@@ -22,28 +24,33 @@ public class VerificationService implements com.sinch.sdk.domains.verification.V
...
+  private Map<String, AuthManager> authManagers;
 
   public VerificationService(Configuration configuration, HttpClient httpClient) {
     this.configuration = configuration;
     this.httpClient = httpClient;
 
+    // by default, unified id/secret from configuration are used but can be super sed if
+    // "useSecrets"
+    // is called after initialization
+    useSecrets(
```
asein-sinch: Sorry to come back on this, but I have 2 remarks on this method:
 - its name: `useSecrets` doesn't reflect the reality of the required credentials. When the developer will use this method, he will give the application key and the application secret from an application created on the verification or voice dashboard. Hence I would suggest to change the name to: `useApplication` or `useApplicationCredentials`
 - its parameters: as explained above, the parameters will be collected from the verification or voice dashboard. When copying the parameters, the application secret is displayed in its base64-encoded value, so on lines 38-39, you are encoding it again which is not correct. The information about the value being encoded is necessary only for the server receiving the signed request, in order to calculate the signature with the same secret as the client.
There is no need to think forward about what will happen when we will have the "unified signing experience" with OAuth2 as the current solution only defines access keys to segregate your use cases, but doesn't offer the level of configuration found in the voice and verification apps. If apps can be defined under an access key, then it will be all right... we'll see!

JPPortier: - replaced by `setApplicationCredentials` to stay with the setter pattern used by SDK
- removed the default configuration usage based onto unified credentials because of not yet available (to avoid misunderstanding about what is supported and inputs)
- added a check against undefined credentials before calling underlying service to have exception message from SDK and not from server

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationService.java
```diff
@@ -22,45 +24,55 @@ public class VerificationService implements com.sinch.sdk.domains.verification.V
...
+              configuration, httpClient, authManagersSupplier);
     }
     return this.status;
   }
+
+  private void checkCredentials() throws ApiAuthException {
+    if (null == authManagers || authManagers.isEmpty()) {
+      throw new ApiAuthException(
+          String.format(
+              "Service '%s' cannot be called without defined credentials",
```
asein-sinch, line 74: Maybe we can be more explicit about what are the "defined credentials" and give information about which method to call

client/src/main/com/sinch/sdk/domains/verification/adapters/StatusService.java
```diff
@@ -14,22 +14,29 @@
...
+      Supplier<Map<String, AuthManager>> authManagerSupplier) {
+    this.configuration = configuration;
+    this.httpClient = httpClient;
+    this.authManagerSupplier = authManagerSupplier;
   }
 
-  private QueryVerificationsApi getApi() {
-    return this.api;
+  protected QueryVerificationsApi getApi() {
+    return new QueryVerificationsApi(
```
asein-sinch, line 35: I've noticed that you have changed the way you initialize the api; it's no more a class property.
Is it because you need to call the authManagerSupplier in case the application credential have changed?

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationService.java
JPPortier: Yes, but keep in mind this is not yet the code which will be released
This feature branch is for Verification API support and I do not plan to release this branch to main until we did met an agreement about current ADR related to auth aspect.
So what is present here is to be able to going forward for Verification API endpoints support but will have to be review in regard of authentication mechanism before final feature release

client/src/main/com/sinch/sdk/domains/verification/adapters/StatusService.java
JPPortier: Exact, now the credentials are more dynamic than static; but this far to be perfect in context of Immutable pattern (see current ADR comments and previous comment because this cannot be the final implementation)

asein-sinch: All right, let's keep it this way for the moment then

asein-sinch (approved):

PR #21, opened 2023-11-28
DEVEXP-111: add User-Agent SDK header

JPPortier:
Sending SDK User-Agent header.
Format will be: `sinch-sdk/{sdk_version} ({language}/{language_version}; {implementation_type}; {auxiliary_flag})`

Where:
- `sdk_version`: Version according to release/deployment version
- `language`: hard coded to `Java`
- `language_version`: java compiler version set for sources
- `implementation_type`: Currently only `Apache` for HTTP Apache client is supported
- `auxiliary_flag`: Comma separated list containing information about JVM used for SinchClient execution:
   - `java.vendor`  
   - `java.version`

Example of generated header: `User-Agent: sinch-sdk/0.0.1-SNAPSHOT (Java/1.8; Apache; AdoptOpenJDK,14.0.2)`

### DIFF
--- a/client/resources/version.properties
+++ b/client/resources/version.properties
@@ -1,0 +1,4 @@
+project.version=${project.version}
+project.name=${project.name}
+project.auxiliary_flag=
+project.java.version=${maven.compiler.source}
--- a/client/resources/version.properties
+++ b/client/resources/version.properties
@@ -1,4 +1,3 @@
 project.version=${project.version}
 project.name=${project.name}
 project.auxiliary_flag=
-project.java.version=${maven.compiler.source}
--- a/client/src/main/com/sinch/sdk/SinchClient.java
+++ b/client/src/main/com/sinch/sdk/SinchClient.java
@@ -35,7 +35,6 @@ public class SinchClient {
 
   private static final String PROJECT_NAME_KEY = "project.name";
   private static final String PROJECT_VERSION_KEY = "project.version";
-  private static final String PROJECT_JAVA_VERSION_KEY = "project.java.version";
   private static final String PROJECT_AUXILIARY_FLAG = "project.auxiliary_flag";
 
   // sinch-sdk/{sdk_version} ({language}/{language_version}; {implementation_type};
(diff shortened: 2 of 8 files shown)


client/resources/version.properties
```diff
@@ -0,0 +1,4 @@
+project.version=${project.version}
+project.name=${project.name}
+project.auxiliary_flag=
+project.java.version=${maven.compiler.source}
```
asein-sinch: The ADR says that this field is used for "Version of the language used for executing the SDK". Can you confirm?

JPPortier: I asked myself same kind of question...
My reading of "language_version" was "language version" which is currently very static 😕 (but in the future we could imaging releasing a multiple release JAR enabling to use java11 native http client)
And this is why I added the JVM related information which is very dynamic...
@650elx do you have some feedback ?

650elx: `language_version` purpose is to gather statistical data about what `language versions` customers are using.
For example; PythonSDK supports version 3.8, 3.9, 3.10, 3.11 - one of those will end up in header.

Node JS?
Let's say we support two versions; 18LTS and 20LTS - one of those will end up in header.

Why we need that information?
To know what's import for customers so we can make better business decisions.

Same with `implementation type` or `auxiliary_flag`

JPPortier: Ok, thank you.
For java we cannot have access to the "language versions customers are using"
We can have access to :
- what was used during compilation for Sinch Java SDK (this is the project.java.version=${maven.compiler.source} I'm sending to header
- JVM version used for execution

None of these are related to client java version they are using.
Even JVM version is not related to language version: you can compile for java 11 and use a java 17 JVM instance.

So, @asein-sinch , in current state, I think current header is fine; and in the future if we support multi release jar, then, the java version could have more values

650elx: @JPPortier So maybe `JVM version used for execution` is what we're looking for?

@Dovchik How the thing looks like on the .NET side?

JPPortier: > @JPPortier So maybe `JVM version used for execution` is what we're looking for?

@650elx This why these information are present within `auxilliary_flag`
From sample above; `...; AdoptOpenJDK,14.0.2)`

Dovchik: It,s straightforward version of NET
Monickers like: net5, net6, net7, net4.8, netcore3

JPPortier: @asein-sinch @650elx 
Updated according to last synch: 
- the `language_version` is based onto `JVM version`
- `auxilliary_flag` contains the  `JVM name`

Example: `User-Agent: sinch-sdk/0.0.1-SNAPSHOT (Java/17.0.8; Apache; Eclipse Adoptium)`

asein-sinch (approved):

PR #25, opened 2023-11-29
Create integrated numbers flow

JPPortier:
As discussed with Jean-Pierre, I put myself in the shoes of a SDK user and wrote a little program to use some of the Numbers APIs


core/src/main/com/sinch/sdk/core/models/pagination/ListResponse.java
```diff
@@ -50,6 +52,19 @@ public Iterator<T> iterator() {
...
 
+  /**
+   * Getting a stream across all items
+   *
+   * <p>Underline API (HTTP request) will be called on demand when next page content will be
+   * required to fulfill iterator with items for consecutive page
+   *
+   * @return Stream onto items
+   */
+  public Stream<T> stream() {
```
asein-sinch, line 63: Nice! 🌟

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
```diff
@@ -0,0 +1,392 @@
...
+            + " part of them");
+
+    // 1. Build the request data for listing the regions that support 'LOCAL' numbers
+    var availableRegionsRequestData =
+        AvailableRegionListAllRequestParameters.builder()
+            .setTypes(Collections.singletonList(NumberType.LOCAL))
+            .build();
+
+    // 2. Request for available regions using the built-in SDK method
+    AvailableRegionListResponse availableRegionListResponse =
```
asein-sinch: We can use a `var` for the response too (in several places in this file)

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
```diff
@@ -0,0 +1,392 @@
...
+                NumberPattern.builder()
+                    .setSearchPattern(SearchPattern.START)
+                    .setPattern(pattern)
+                    .build())
+            .build();
+
+    // 2. rent a number
+    ActiveNumber rentedNumber = sinchClient.numbers().available().rentAny(rentAnyNumberRequestData);
+
+    echo(
```
asein-sinch, line 218: You may also add the other log `echo("Let's use now the 'ActiveNumbers' API to see how to manage this number.\n");` to have the same output with `rent` and `rentAny`

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
```diff
@@ -0,0 +1,392 @@
...
+          activeNumbersListResponse.getContent().stream()
+              .filter(
+                  f ->
+                      phoneNumber.equals(f.getPhoneNumber())
+                          && regionCode.equals(f.getRegionCode()))
+              .findFirst();
+      if (found.isEmpty() && activeNumbersListResponse.hasNextPage()) {
+        activeNumbersListResponse = activeNumbersListResponse.nextPage();
+      } else {
+        activeNumbersListResponse = null;
```
asein-sinch: This looks like a code smell. Can we check if we have reached the end of the list instead?

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
```diff
@@ -0,0 +1,392 @@
...
+    // 1. Build the request data
+    var listActiveNumbersRequestData =
+        ActiveNumberListRequestParameters.builder()
+            .setRegionCode(regionCode)
+            .setType(NumberType.LOCAL)
+            .setPageSize(2) // Set a low number to demonstrate the pagination later on
+            .build();
+
+    // 2. Request for active number using the built-in SDK method
+    ActiveNumberListResponse activeNumbersListResponse =
```
asein-sinch: The steps 1 and 2 are duplicated from `verifyActiveNumberByPagination`. I understand we need to "reset" the activeNumbersListResponse in order to run the iterator from the first elements returned by the server.
But in this example, it's maybe clearer to have a duplication rather to try to factorise this part

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
```diff
@@ -0,0 +1,392 @@
...
+
+    return phoneNumber;
+  }
+
+  void rentPhoneNumber(int step, SinchClient sinchClient, String phoneNumber) {
+
+    echoStep(step, "Rent phone number: '" + phoneNumber + "'");
+
+    // 1. Build the request data
+    var rentRequestData = AvailableNumberRentRequestParameters.builder().build();
```
asein-sinch: Here we are not mentioning any sms or voice configuration. They are optional in the OAS documentation, but maybe it doesn't make sense to rent a number without any of these configuration (unless we update the active number accordingly in a subsequent API call)

asein-sinch (dismissed):

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
JPPortier: fixed

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
JPPortier: Right. Moved this log at upper level

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
JPPortier: The end of the list (pages) is when we reached hasNextPage set to false.
Changed to improve code readability

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
JPPortier: Right; ideally the `sinchClient.numbers().available().rent` should accept a `null` parameter or provide a second function without the `parameters` parameter.

I modified the `rent` function to accept null parameter but changed the sample too, to set a callback URL parameter: this is here the purpose of the sample: an on-boarding experience

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
JPPortier: Will synch; need to clarify what you mean

asein-sinch (approved):

PR #26, opened 2023-12-04
DEVEXP-201: Verification webhooks

JPPortier:
This PR is adding helpers form SDK for webhooks support to: 
- `checkAuthentication`: validate authentication has received from request. User do not have to write it own implementation related to hashcode computation/validation 
- `unserializeVerificationEvent` && `serializeVerificationResponse` serialize/de-serialize payloads for webhooks callbacks and provide to user ready-to-use classes instance. So do not have to implement it owns outside of SDK
- a ready to use SpringBoot application as sample for a webhooks backend implementation
- samples & code cleanup

To provide these helpers, we need to be able to handle both Authentication validation mechanism (`Basic` and `Application`) because this setting is managed by user from his dashboard.
So even if we only use `Application` authentication when SDK is sending request; webhooks validation handle both of them on received requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Java SDK
 [![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](https://github.com/sinch/sinch-sdk-python/blob/main/LICENSE)
 
 
-[![Python 3.8](https://img.shields.io/badge/Java-8-blue.svg)](https://docs.oracle.com/javase/8)
+[![Java 8](https://img.shields.io/badge/Java-8-blue.svg)](https://docs.oracle.com/javase/8)
 
 
 </h1>
@@ -34,7 +34,6 @@ For more in depth version of the Sinch APIs, please refer to the official develo
 
 - JDK 8 or later
 - [Maven](https://maven.apache.org/)
-- [Maven Central](https://mvnrepository.com/artifact/com.sinch.sdk/sinch-java-sdk)
 - [Sinch account](https://dashboard.sinch.com)
 
 ## Installation
--- a/client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
+++ b/client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
@@ -11,7 +11,6 @@ import java.nio.charset.StandardCharsets;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.time.Instant;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collection;
@@ -42,7 +41,7 @@ public class VerificationApplicationAuthManager implements AuthManager {
 
   @Override
   public Collection<Pair<String, String>> getAuthorizationHeaders(
-      String method, String httpContentType, String path, String body) {
+      String timestamp, String method, String httpContentType, String path, String body) {
 
     String decodePath;
     try {
@@ -52,7 +51,6 @@ public class VerificationApplicationAuthManager implements AuthManager {
     }
     // see
     // https://developers.sinch.com/docs/verification/api-reference/authentication/signed-request/
-    Instant timestamp = Instant.now();
     String bodyMD5Hash = getBodyMD5Hash(body);
     String stringToSign = getSignature(method, bodyMD5Hash, httpContentType, timestamp, decodePath);
     String encoded = encode(stringToSign);
@@ -78,13 +76,13 @@ public class VerificationApplicationAuthManager implements AuthManager {
   }
 
   private String getSignature(
-      String method, String bodyMD5Hash, String httpContentType, Instant timestamp, String path) {
+      String method, String bodyMD5Hash, String httpContentType, String timestamp, String path) {
     return String.join(
         "\n",
         method,
         bodyMD5Hash,
         null != httpContentType ? httpContentType : "",
-        XTIMESTAMP_HEADER + ":" + timestamp.toString(),
+        XTIMESTAMP_HEADER + ":" + timestamp,
         path);
   }
 
--- a/client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
+++ b/client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
@@ -58,7 +58,7 @@ public class VerificationApplicationAuthManager implements AuthManager {
 
     return Arrays.asList(
         new Pair<>("Authorization", AUTH_KEYWORD + " " + key + ":" + encoded),
-        new Pair<>(XTIMESTAMP_HEADER, timestamp.toString()));
+        new Pair<>(XTIMESTAMP_HEADER, timestamp));
   }
 
   private String getBodyMD5Hash(String body) {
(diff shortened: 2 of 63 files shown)


JPPortier:
Please ignore failures for samples
Will fix it, they are due to mix between java used  versions but samples are running fine

README.md
```diff
@@ -10,7 +10,7 @@ Java SDK
 [![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](https://github.com/sinch/sinch-sdk-python/blob/main/LICENSE)
 
 
-[![Python 3.8](https://img.shields.io/badge/Java-8-blue.svg)](https://docs.oracle.com/javase/8)
```
650elx, line 13: That's an interesting one :D

650elx (approved):
LGTM

client/src/main/com/sinch/sdk/domains/verification/WebHooksService.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param method      The HTTP method used ot handle the callback
+   * @param path        The path to you backend endpoint used for callback
+   * @param headers     Received headers
+   * @param jsonPayload Received payload
+   * @return Is authentication is validated (true) or not (false)
+   *
+   * see <a href="https://developers.sinch.com/docs/verification/api-reference/authentication/callback-signed-request">https://developers.sinch.com/docs/verification/api-reference/authentication/callback-signed-request</a>
+   * @since 1.0
+   */
+  boolean checkAuthentication(
```
asein-sinch, line 37: I'm wondering if it makes sense to have this method in this service, knowing that the voice domain will define the same

client/src/main/com/sinch/sdk/domains/verification/adapters/WebHooksService.java
```diff
@@ -0,0 +1,89 @@
...
+    String computedAuthorization =
+        computedHeaders.stream()
+            .filter(f -> f.getLeft().equals("Authorization"))
+            .findFirst()
+            .map(Pair::getRight)
+            .orElse("");
+    String[] newSplit = computedAuthorization.split(" ");
+    String computedHash = newSplit.length > 1 ? newSplit[1] : "";
+
+    return computedHash.equals(authorizationHash);
```
asein-sinch: This function may be refactored to just be about the comparison between the received against the computed value and put the implementation details about how the extract the data in a private function

client/src/test/java/com/sinch/sdk/domains/verification/adapters/WebhooksServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    Assertions.assertThat(authenticationResult).isEqualTo(true);
+  }
+
+  @Test
+  void checkApplicationAuthenticationFailure() throws ApiException {
+
+    Map<String, String> headers =
+        Stream.of(
+                new AbstractMap.SimpleEntry<>("authorization", "application foo="),
```
asein-sinch: Maybe you can test against a value that starts with the application key but with a bad signature only such as `789:foo=` so that you are really checking the tampering of the payload

client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
```diff
@@ -42,7 +41,7 @@ public void resetToken() {
 
   @Override
   public Collection<Pair<String, String>> getAuthorizationHeaders(
-      String method, String httpContentType, String path, String body) {
+      String timestamp, String method, String httpContentType, String path, String body) {
```
asein-sinch, line 44: You may remove the `toString()` when setting the x-timestamp header at line https://github.com/sinch/sinch-sdk-java/pull/26/files#diff-feb9f88279fca9ebde2e5c47af9da62694ad1d41b00f3e6f5f3672d6b51e3fa2L63

core/src/main/com/sinch/sdk/core/http/AuthManager.java
```diff
@@ -13,6 +14,11 @@ public interface AuthManager {
 
   void resetToken();
 
+  default Collection<Pair<String, String>> getAuthorizationHeaders(
+      String method, String httpContentType, String path, String body) {
+    return getAuthorizationHeaders(Instant.now().toString(), method, httpContentType, path, body);
```
asein-sinch, line 19: Why do you define the timestamp at this level?
The other parameters are static, coming from the specification. This one is dynamic at could be generated in the `VerificationApplicationAuthManager` as it was before.

sample-app/src/main/java/com/sinch/sample/webhooks/verification/VerificationController.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // decode the request payload
+    var event = sinchClient.verification().webhooks().unserializeVerificationEvent(body);
+
+    // let business layer process the request
+    var response = switch (event) {
+      case VerificationRequestEvent e -> service.verificationEvent(e);
+      case VerificationResultEvent e -> {
+        service.verificationEvent(e);
```
asein-sinch, line 67: I was confused by the fact you were calling the same method as in the previous `case` but with another object... until I found the so tiny implementation!

client/src/main/com/sinch/sdk/domains/verification/WebHooksService.java
JPPortier: Right.
When Voice will arise, it seems refactoring will have sense (something synch with what we have to do related to Application spreading across Voice/Verification)

client/src/main/com/sinch/sdk/domains/verification/adapters/WebHooksService.java
JPPortier: Done.

client/src/test/java/com/sinch/sdk/domains/verification/adapters/WebhooksServiceTest.java
JPPortier: Tests added

client/src/main/com/sinch/sdk/auth/adapters/VerificationApplicationAuthManager.java
JPPortier: Thanks
Fixed

core/src/main/com/sinch/sdk/core/http/AuthManager.java
JPPortier: In fact this exactly the purpose of what you're pointing: this function is a default for an interface.
So it is a helper for services calling VerificationApplicationAuthManager (currently the HttpApacheClient): they do not have to provide the timestamp value on call.
In case of others http client implementation come, they could avoid the knowledge of this timestamp usage; and the helper is shared by all of them.

By opposition, the second implementation (getAuthorizationHeaders with timeStamp as String parameter) is having to be implemented by concrete classes and will be called from webhook service; because the webhooks service is having to passthrough the timestamp value, received as String for computation and check with the recevied hash.

sample-app/src/main/java/com/sinch/sample/webhooks/verification/VerificationController.java
JPPortier: Java 21 inside 😄

core/src/main/com/sinch/sdk/core/http/AuthManager.java
asein-sinch: Ahhh I missed this point. Very nice!

asein-sinch (approved):
Very nice ⭐️

PR #27, opened 2023-12-04
DEVEXP-201: store application credentials at SinchClient configuration level

JPPortier:
…hey are store at SinchClient level and a SinchClient instance can handle a single one application (like for single one project id)


650elx (approved):
👌

PR #30, opened 2023-12-12
doc: README update

JPPortier:
Update README markdown according to documentation template


README.md
```diff
@@ -1,115 +1,107 @@
...
\ No newline at end of file
+# Sinch Java SDK
+
+Here you'll find documentation related to the Sinch Java SDK, including how to install it, initialize it, and start developing <language> code using Sinch services.
+
+To use Sinch services, you'll need a Sinch account and access keys. You can sign up for an account and create access keys at [dashboard.sinch.com](https://dashboard.sinch.com).
+
+For more information on the Sinch APIs on which this SDK is based, refer to the official [developer documentation portal](https://developers.sinch.com).
+
+**This SDK is currently available to selected developers for preview use only. It is being provided for the purpose of collecting feedback, and should not be used in production environments.**
```
Dovchik: Side note, who is the `selected developers`? Can we just mention that it's prerelease so prod use is not recommended and a possibility of breaking changes to come?

Dovchik (dismissed):

README.md
JPPortier: @alex-sberna : What do you think about ?

alex-sberna: I think this is a good idea. We can probably even just copy the same note that's on developers.sinch.com: "This SDK is currently available for preview purposes only. It should not be used in production environments."

JPPortier:
```suggestion
**This SDK is currently available for preview purposes only. It should not be used in production environments.**
```

Dovchik (approved):

PR #33, opened 2023-12-19
feat (DEVEXP-148): Voice Callouts

JPPortier:


client/src/main/com/sinch/sdk/auth/adapters/VoiceApplicationAuthManager.java
```diff
@@ -0,0 +1,9 @@
+package com.sinch.sdk.auth.adapters;
+
+// Voice and Verification are sharing same authentication mechanism
+public class VoiceApplicationAuthManager extends VerificationApplicationAuthManager {
```
asein-sinch: Here we have `VoiceApplicationAuthManager` that extends `VerificationApplicationAuthManager ` because you have created the SDK for these APIs in this order, and it could have been the opposite. Do you think it makes sense to have a single `ApplicationAuthManager` for both APIs ?

JPPortier: Yes totally
credentials are already under a cross domains "application" term
Will do it

client/src/main/com/sinch/sdk/domains/voice/adapters/VoiceService.java
```diff
@@ -0,0 +1,63 @@
...
+  private final HttpClient httpClient;
+  private CalloutsService callouts;
+
+  private Map<String, AuthManager> clientAuthManagers;
+
+  public VoiceService(Configuration configuration, HttpClient httpClient) {
+
+    // Currently, we are not supporting unified credentials: ensure application credentials are
+    // defined
+    Objects.requireNonNull(configuration.getApplicationKey(), "'applicationKey' cannot be null");
```
asein-sinch, line 27: Maybe we can also check the application key and secret are not empty strings

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
```diff
@@ -0,0 +1,128 @@
...
+    client.getEnableAce().ifPresent(dto::setEnableAce);
+    client.getEnableDice().ifPresent(dto::setEnableDice);
+    client.getEnablePie().ifPresent(dto::setEnablePie);
+    client.getLocale().ifPresent(dto::setLocale);
+    client.getGreeting().ifPresent(dto::setGreeting);
+    client.getMohClass().ifPresent(f -> dto.setMohClass(f.value()));
+    client.getDomain().ifPresent(f -> dto.setDomain(f.value()));
+
+    return new CalloutRequestDto()
+        .method(MethodEnum.CONFERENCECALLOUT.getValue())
```
asein-sinch, line 67: Is this enum name generated from the spec? Otherwise, you may rename it `CONFERENCE_CALLOUT`

client/src/main/com/sinch/sdk/domains/voice/adapters/VoiceService.java
JPPortier: Is it really useful ?
Next step could not null and nor empty and nor empty when spaces or dead characters are stripped from the string.
In this case checking against a null value we are assuming developers had set a value.... and we could expect that the contract is satisfied.

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
```diff
@@ -0,0 +1,128 @@
...
+    client.getCustom().ifPresent(dto::setCustom);
+    client.getEnableAce().ifPresent(dto::setEnableAce);
+    client.getEnableDice().ifPresent(dto::setEnableDice);
+    client.getEnablePie().ifPresent(dto::setEnablePie);
+    client.getLocale().ifPresent(dto::setLocale);
+    client.getDomain().ifPresent(f -> dto.setDomain(f.value()));
+    client.getText().ifPresent(dto::setText);
+    client.getPrompts().ifPresent(dto::setPrompts);
+
+    return new CalloutRequestDto().method(MethodEnum.TTSCALLOUT.getValue()).ttsCallout(dto);
```
asein-sinch, line 87: Same comment for `TTSCALLOUT`

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
```diff
@@ -0,0 +1,128 @@
...
+    client.getCli().ifPresent(f -> dto.setCli(f.stringValue()));
+    client.getDtfm().ifPresent(dto::setDtmf);
+    client.getCustom().ifPresent(dto::setCustom);
+
+    client.getMaxDuration().ifPresent(dto::setMaxDuration);
+    client.getIce().ifPresent(dto::setIce);
+    client.getAce().ifPresent(dto::setAce);
+    client.getPie().ifPresent(dto::setPie);
+
+    return new CalloutRequestDto().method(MethodEnum.CUSTOMCALLOUT.getValue()).customCallout(dto);
```
asein-sinch, line 104: And for `CUSTOMCALLOUT`

client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -57,4 +59,48 @@ void defaultSmsUrlAvailable() {
     SinchClient client = new SinchClient(configuration);
     assertNotNull(client.getConfiguration().getSmsUrl());
   }
+
+  @Test
+  void defaultVerificationUrlAvailable() {
+    Configuration configuration =
+        Configuration.builder().setKeyId("foo").setKeySecret("foo").setProjectId("foo").build();
+    SinchClient client = new SinchClient(configuration);
+    assertNotNull(client.getConfiguration().getVerificationUrl());
```
asein-sinch, line 68: Instead of checking for not null, you can check for the real value. For the verification API, it's fine as there is only one URL, but for voice, you may want to check it uses the default one

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
JPPortier: It come from generated file from OAS: https://github.com/sinch/sinch-sdk-java/blob/feat/voice-callout/openapi-contracts/src/main/com/sinch/sdk/domains/voice/models/dto/v1/CalloutRequestDto.java#L37-L43

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
JPPortier: _Same answer as for CONFERENCE_CALLOUT_

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
JPPortier: _Same answer as for CONFERENCE_CALLOUT_

client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -57,4 +59,48 @@ void defaultSmsUrlAvailable() {
...
+  }
+
+  @Test
+  void defaultVoiceUrl() {
+    Configuration configuration =
+        Configuration.builder().setKeyId("foo").setKeySecret("foo").setProjectId("foo").build();
```
asein-sinch, line 74: About the credentials management, which direction did you choose finally? Because here I see that the project Id + access key and secret are used to initialize the configuration object although these credentials are not required for using the voice API alone

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: The test is against checking a default value is present
The `defaultVoiceUrl` test is checking against a not empty value
`voiceUrlFromRegion` test is the only we are checking against a value coming from configuration file. Ideally we should set for test such value to not depend of configuration values
So, back to the current test, this is why it is not checking against a hard coded value... it is not good enough

client/src/test/java/com/sinch/sdk/models/ConfigurationTest.java
```diff
@@ -22,6 +25,8 @@ class ConfigurationTest {
           .setNumbersUrl(NUMBERS_SERVER)
           .setSmsRegion(SMS_REGION)
           .setSmsUrl(SMS_SERVER)
+          .setApplicationKey(APPLICATION_KEY)
+          .setApplicationSecret(APPLICATION_SECRET)
           .build();
 
   @Test
```
asein-sinch, line 32: I took me a while to remember that the credential values should not appear as a result of the object conversion to a string. Do you think it's worth to rename the method such as `testToStringShouldNotDiscloseCredentialValues()` ?

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: Yes.
Currently waiting for ADR validation for credentials
We do not have choice to going to be less restrictive (credentials by domains) just we have to be consistent across SDKs

client/src/main/com/sinch/sdk/domains/voice/adapters/VoiceService.java
asein-sinch: Indeed, I was wondering if we could get an glimpse at the application key generator to determine the format, as we do for credit card numbers... but you're right, let's assume that when a developer provides a non-null value, he provides the correct one. If not, the server will be pleased to tell him the credentials are not correct ;)

client/src/test/java/com/sinch/sdk/models/ConfigurationTest.java
JPPortier: Don't the assert error messages was not enough from `testToString` to remind this ?

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
asein-sinch: /me crying silently

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
asein-sinch: /me sobbing silently

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
asein-sinch: /me drying my tears and facing the hard truth

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/CalloutsDtoConverter.java
JPPortier: OAS file: 
```
          enum:
            - conferenceCallout
            - ttsCallout
            - customCallout
```
So let say the generator just do the job...

client/src/test/java/com/sinch/sdk/SinchClientTest.java
asein-sinch: Ok, let's wait for the meeting on Jan 10th then. Wow that's next year!!

client/src/test/java/com/sinch/sdk/SinchClientTest.java
asein-sinch: Ok, it's true you are using a configuration file. On my side, I've chosen to hardcode the URL from the specification and to provide a default value, but I understand your point :)

client/src/test/java/com/sinch/sdk/models/ConfigurationTest.java
asein-sinch: With the method name, I was expecting that you would test the `toString()` against the expected result. This is why I was puzzled that you were doing some assertFalse. That's up to you ;)

client/src/main/com/sinch/sdk/auth/adapters/VoiceApplicationAuthManager.java
asein-sinch: Nice, do you want to do it in this PR or open another?

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: Yes configuration file store all possible URLS according to possible defined regions.

It is still possible to set manually another URL outside of predefined region (e2e scope) to bypass default value coming from "global" region

client/src/main/com/sinch/sdk/auth/adapters/VoiceApplicationAuthManager.java
JPPortier: No, will use this PR

client/src/test/java/com/sinch/sdk/SinchClientTest.java
asein-sinch: Let's add memes in our PRs :)

image

client/src/test/java/com/sinch/sdk/models/ConfigurationTest.java
JPPortier: Yes; I understand your point of view.... and this is why the point of view to test what toString is doing is having sense... just to ensure it do not output too much information
To be honest; checking the output of toString  for another context could seems to be over-engineering

client/src/main/com/sinch/sdk/auth/adapters/VoiceApplicationAuthManager.java
JPPortier: Done

asein-sinch (approved):
⭐️

PR #36, opened 2024-01-03
Feat (DEVEXP-213): Voice API / Conferences

JPPortier:


JPPortier:
Will rebase branch onto main (support capability to handle `undefined` vs `null` values for request)

client/src/main/com/sinch/sdk/domains/voice/adapters/CalloutsService.java
```diff
@@ -23,6 +26,21 @@ protected CalloutsApi getApi() {
...
+
+    return call(parameters);
+  }
+
+  public String custom(CalloutRequestParametersCustom parameters) {
+
+    return call(parameters);
+  }
+
   public String call(CalloutRequestParameters parameters) {
```
asein-sinch, line 44: Do you still want to expose the generic function?

client/src/main/com/sinch/sdk/domains/voice/adapters/ConferencesService.java
```diff
@@ -0,0 +1,54 @@
...
+            httpClient, configuration.getVoiceServer(), authManagers, new HttpMapper());
+  }
+
+  protected ConferencesApi getApi() {
+    return this.api;
+  }
+
+  public String call(CalloutRequestParametersConference parameters) {
+    return CalloutsDtoConverter.convert(
+        getApi().callouts(CalloutsDtoConverter.convert(parameters)));
```
asein-sinch, line 33: What's the final word about the this? Do we need to expose the same service through 2 different paths?

JPPortier: Not sure there was discussion about this point.
Because of documentation and OAS tags the service twice, I just added dedicated call to this service

client/src/main/com/sinch/sdk/domains/voice/adapters/CalloutsService.java
JPPortier: Yes: My thought was to let it public so if final end user choose to store the parameters as `CalloutRequestParameters` "base class" he can call this function.
Following use case may not have sense, to be confirmed, but technically a user could prepare a list of calls and store parameters within a `Collection<CalloutRequestParameters>`
So by batch, or any use case, it could call `call(CalloutRequestParameters parameters)`and not any of dedicated functions

sample-app/src/main/java/com/sinch/sample/numbers/NumbersSampleFlow.java
```diff
@@ -377,17 +380,4 @@ void releasePhoneNumber(int step, SinchClient sinchClient, String phoneNumber) {
             + " has been released; it will expire at "
             + releasedNumber.getExpireAt());
   }
-
-  void echoStep(int step, String text) {
```
asein-sinch, line 381: 👍

sample-app/src/main/java/com/sinch/sample/voice/conferences/KickParticipant.java
```diff
@@ -0,0 +1,29 @@
...
+      new KickParticipant().run();
+    } catch (Exception e) {
+      LOGGER.severe(e.getMessage());
+      e.printStackTrace();
+    }
+  }
+
+  public void run() {
+
+    String callId = "e429ecbb-6057-4756-8148-2b498f968136";
```
asein-sinch, line 24: This data should come from a config file, right?

sample-app/src/main/java/com/sinch/sample/voice/ConferencesSampleFlow.java
asein-sinch, line 1: Congrats for putting together a scenario with the Voice API 👏

sample-app/src/main/java/com/sinch/sample/voice/conferences/KickParticipant.java
JPPortier: Right, it could but because of very volatile I did not handle it from config
I'll do for consistency (will change it with incoming work related to Voice API)

asein-sinch (approved):

PR #37, opened 2024-01-05
DEVEXP-264: validate E164 number format

JPPortier:


client/src/main/com/sinch/sdk/models/E164PhoneNumber.java
```diff
@@ -1,11 +1,20 @@
 package com.sinch.sdk.models;
 
 import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class E164PhoneNumber {
+
+  private static final Pattern pattern = Pattern.compile("^\\+(?:[0-9] ?){6,14}[0-9]$");
```
asein-sinch: The naming convention for a private static final object would be `PATTERN`. May you can rename it to E164_PATTERN.
Also, I'm not sure the regexp is valid as a country code cannot start with 0.
Moreover, the pattern would be different depending on the country and even if a number is matching the pattern, it can still be invalid, such as +330000000000

JPPortier: `pattern` -> `PATTERN` : 👍 
country code and zero: will investigate
About pattern different according to country, I was not thinking going to the details level. Aim was just about E164 format validation but not (yet) validation by country

JPPortier: Updated:
- pattern -> PATTERN
- take into account invalid region code starting with zero
- by choice: do not invalid number due to specific region/country. It could be a further enhancement

Because of server is accepting strings like `+330123456..` but also `+33 0123456..`, `+(33) 0123456..` these cases are also valid cases from client side

Dovchik: On what source your pattern is based? Update: discard, I saw a comment in PR

Dovchik (approved):

asein-sinch (approved):
It would be good to check the code on the server side in order to perform the same validation: if we miss one valid pattern (ie accepted by the server), this model would limit the SDK usage compared to a direct API call

JPPortier:
> It would be good to check the code on the server side in order to perform the same validation: if we miss one valid pattern (ie accepted by the server), this model would limit the SDK usage compared to a direct API call

I agree.
I'm still waiting for these feedback and because this validation if for a feature branch, not yet released, this is not yet a blocker
In addition, I'm planning to support a flag to have a "relax" vs "strict" (the current and default) behavior. So this will enable to be confident in case of "too strict validation"

PR #38, opened 2024-01-11
DEVEXP-212: Voice/Calls

JPPortier:
- Support Voice.calls APIS
- DualToneMultiFrequency (DTMF) format validation client side
- E164Number and DualToneMultiFrequency format validation can be enabled/disabled programmatically (relax/strict mode. `strict` is the default)

Note: javadoc for SVAML classes not finalized. This will be done after feedback from Voice/Documentation team


openapi-contracts/src/main/com/sinch/sdk/domains/voice/models/dto/v1/DestinationTypeDto.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Can be of type &#x60;number&#x60; for PSTN endpoints or of type &#x60;username&#x60; for data
+ * endpoints.
+ */
+public enum DestinationTypeDto {
+  NUMBER("number"),
+
+  NUMBER2("Number"),
```
asein-sinch, line 25: 😭 I'm sure we can find a better name, or maybe convert the value in lower case before trying to match

client/src/main/com/sinch/sdk/models/E164PhoneNumber.java
```diff
@@ -25,9 +25,11 @@ public class E164PhoneNumber {
 
   private final String number;
 
+  private static boolean strict = true;
```
asein-sinch: Are you sure you want to active strict mode by default?
And in case the strict mode is set to false, you may print a warning

client/src/main/com/sinch/sdk/models/DualToneMultiFrequency.java
```diff
@@ -0,0 +1,98 @@
...
+  public static DualToneMultiFrequency valueOf(String value) {
+    return new DualToneMultiFrequency(value);
+  }
+
+  /**
+   * Configure if invalid string values will throw an Exception or not
+   *
+   * @param strict Set strict mode to true/false
+   */
+  public static void setStrict(boolean strict) {
```
asein-sinch, line 60: It seems this method is not thread safe 🙃

openapi-contracts/src/main/com/sinch/sdk/domains/voice/models/dto/v1/SvamlActionConnectConfDto.java
```diff
@@ -32,6 +34,39 @@
 @JsonFilter("uninitializedFilter")
 @JsonInclude(value = JsonInclude.Include.CUSTOM)
 public class SvamlActionConnectConfDto {
+  /** The name property. Must have the value &#x60;connectConf&#x60;. */
+  public enum NameEnum {
```
asein-sinch, line 38: Why do you need this enum? Is it because you have zero-trust to the server?

openapi-contracts/src/main/com/sinch/sdk/domains/voice/models/dto/v1/DestinationTypeDto.java
JPPortier: I agree but these files are coming from the open api generator tools so, currently,  this is how it generate names and at mustache template level we do not a lot of possibility to change it.
On the other side, we cannot lower/upper too much at runtime: if server is accepting lowercase as input and return uppercase in some other situation. This is having to be handled from context execution (this is the adapter layers responsibility)

client/src/main/com/sinch/sdk/models/E164PhoneNumber.java
JPPortier: Good point for default value.
For alpha and feedback releases expectation I like the idea to have default to true. But I'm open to change my mind 
Thank you for the message warning: added

client/src/main/com/sinch/sdk/models/DualToneMultiFrequency.java
JPPortier: Exact :-(
it is a static and global variable: aim is if end user is having issue this will have to be set a global level on "application" startup
But you're absolutely right: changed to be thread safe (also done for E164PhoneNumber class)

openapi-contracts/src/main/com/sinch/sdk/domains/voice/models/dto/v1/SvamlActionConnectConfDto.java
JPPortier: Not a zero trust purpose but wish to have a "in specification we trust"
Purpose is to get this value and use value coming the the (OAS) specs and not duplicate expected hard coded value in sources.
Currently the expected value was describe from specs a default value. By adding it as an enum set with a single value; generator create a variable we can use from sources without duplication and hard coded "things"
I anything change from specs about the name: it will break compilation
If change is only for the value: no compilation issue nor at runtime: the expected value will be used without modification from sources

asein-sinch (approved):

PR #42, opened 2024-01-23
DEVEXP-215: SAVML helpers shared by custom callouts requests and webhooks

JPPortier:
refactoring for incoming webhooks, at commit level details:
- Define a 'SAVML Control' object to be used from webhooks and callouts
- Make SAVML parameters optional and define MenuOptionAction helper to define SAVML menu navigation


Dovchik (approved):
LGTM

client/src/main/com/sinch/sdk/domains/voice/models/svaml/MenuOptionAction.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public String getId() {
+    return id;
+  }
+
+  public MenuOptionActionType getType() {
+    return type;
+  }
+
+  public static MenuOptionAction from(MenuOptionActionType type, String id) {
```
asein-sinch, line 25: Usually you are using the builder pattern to construct an Object but here you are using a constructor look-alike. I'm curious about the reason; is it because the 2 parameters are mandatory?

client/src/test/java/com/sinch/sdk/domains/voice/models/requests/CalloutRequestParametersCustomTest.java
```diff
@@ -18,12 +25,25 @@ public class CalloutRequestParametersCustomTest {
           .setDtfm(DualToneMultiFrequency.valueOf("w123#"))
           .setMaxDuration(32)
           .setIce(
-              "{\"action\": {\"name\": \"connectPstn\", \"number\": \"+12233445566\", \"cli\":"
-                  + " \"+12234325234\", \"amd\": {\"enabled\": \"true\"}}}")
+              SVAMLControl.builder()
```
asein-sinch, line 27: OMG, that's so much clearer than an escaped JSON string! Well done!

client/src/test/java/com/sinch/sdk/domains/voice/models/requests/CalloutRequestParametersCustomTest.java
```diff
@@ -18,12 +25,25 @@ public class CalloutRequestParametersCustomTest {
...
+                  .setAction(
+                      ActionConnectPstn.builder()
+                          .setCli("+12234325234")
+                          .setNumber(E164PhoneNumber.valueOf("+12233445566"))
+                          .setAnsweringMachineDetectionEnabled(
+                              AnsweringMachineDetection.builder().setEnabled(true).build())
```
asein-sinch, line 33: I understand the logic to stick to the JSON mapping, but for this particular case, maybe a helper method would be more friendly like `setAnsweringMachineDetectionEnabled(true)` which would act like a DSL

openapi-contracts/src/test/java/com/sinch/sdk/domains/voice/models/dto/svaml/ActionRunMenuDtoTest.java
```diff
@@ -34,7 +34,8 @@ public class ActionRunMenuDtoTest extends BaseTest {
                       .timeoutMills(500)
                       .maxTimeoutMills(123456)
                       .options(
-                          Collections.singletonList(new OptionDto().dtmf("#w").action("menu")))));
+                          Collections.singletonList(
+                              new OptionDto().dtmf("#w").action("menu(foo)")))));
```
asein-sinch, line 38: In the case of an option, I'm not sure `#w` is a valid dtmf as it's intended to be pressed by the user

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/SAVMLActionDtoConverter.java
```diff
@@ -105,15 +106,15 @@ private static SvamlActionConnectPstnDto convertAction(ActionConnectPstn client)
...
-    dto.setAmd(convert(client.getAnsweringMachineDetectionEnabled()));
+    client.getNumber().ifPresent(f -> dto.setNumber(E164PhoneNumberDtoConverter.convert(f)));
+    client.getLocale().ifPresent(dto::setLocale);
+    client.getMaxDuration().ifPresent(dto::setMaxDuration);
+    client.getDialTimeout().ifPresent(dto::setDialTimeout);
+    client.getCli().ifPresent(dto::setCli);
+    client.getSuppressCallbacks().ifPresent(dto::setSuppressCallbacks);
+    client.getDtmf().ifPresent(f -> dto.setDtmf(DualToneMultiFrequencyDtoConverter.convert(f)));
+    client.getIndications().ifPresent(f -> dto.setIndications(EnumDynamicConverter.convert(f)));
+    client.getAnsweringMachineDetectionEnabled().ifPresent(f -> dto.setAmd(convert(f)));
```
asein-sinch: Nitpicking: the setter is an acronym whereas the getter uses the full words

client/src/main/com/sinch/sdk/domains/voice/models/svaml/InstructionSetCookie.java
```diff
@@ -1,20 +1,22 @@
 package com.sinch.sdk.domains.voice.models.svaml;
 
+import com.sinch.sdk.core.models.OptionalValue;
+
 public class InstructionSetCookie extends Instruction {
 
-  private final String key;
-  private final String value;
+  private final OptionalValue<String> key;
+  private final OptionalValue<String> value;
```
asein-sinch: The documentation doesn't specify these properties must not be null, but does it make sense to set a property while the other is not defined?

client/src/main/com/sinch/sdk/domains/voice/models/svaml/InstructionPlayFiles.java
```diff
@@ -1,22 +1,24 @@
 package com.sinch.sdk.domains.voice.models.svaml;
 
+import com.sinch.sdk.core.models.OptionalValue;
 import java.util.Collection;
 
 public class InstructionPlayFiles extends Instruction {
 
-  private final Collection<String> ids;
-  private final String locale;
+  private final OptionalValue<Collection<String>> ids;
```
asein-sinch, line 8: Note for later: maybe we can think about supporting #tts[] and #ssml[]

client/src/main/com/sinch/sdk/domains/voice/models/svaml/ActionConnectSip.java
```diff
@@ -89,46 +90,46 @@ public static Builder<?> builder() {
...
-    Boolean suppressCallbacks;
-    Collection<Pair<String, String>> callheaders;
-    MohClassType moh;
+    OptionalValue<DestinationSip> destination = OptionalValue.empty();
+    OptionalValue<Integer> maxDuration = OptionalValue.empty();
+    OptionalValue<String> cli = OptionalValue.empty();
+    OptionalValue<TransportType> transport = OptionalValue.empty();
+    OptionalValue<Boolean> suppressCallbacks = OptionalValue.empty();
+    OptionalValue<Collection<Pair<String, String>>> callheaders = OptionalValue.empty();
+    OptionalValue<MohClassType> moh = OptionalValue.empty();
```
asein-sinch: Why is the name of this class `MohClassType` and not simply `MohType` like it is with `TransportType` ?

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/SAVMLActionDtoConverter.java
```diff
@@ -240,10 +244,19 @@ private static List<OptionDto> convertMenuOptionCollection(Collection<MenuOption
...
+              return dto;
+            })
         .collect(Collectors.toList());
   }
+
+  private static String convert(MenuOptionAction client) {
+    if (null == client) {
+      return null;
+    }
+    return String.format("%s(%s)", client.getType().value(), client.getId());
```
asein-sinch, line 262: There will be a NPE if the object is built with `MenuOptionAction.from(null, "foo")`

client/src/test/java/com/sinch/sdk/domains/voice/models/requests/CalloutRequestParametersCustomTest.java
asein-sinch: Thinking about it, it's not the complexity of the builder to set a boolean that bothers me the most, it's rather the naming of `setAnsweringMachineDetectionEnabled` which should be `setAnsweringMachineDetection`. The fact that below an object contains a boolean should not be reflected in the upper property according to the spec

client/src/main/com/sinch/sdk/domains/voice/models/svaml/MenuOptionAction.java
JPPortier: Yes, mainly governed by usage simplification

But in the future, if this choice was a bad one; we could move back to builder pattern ... an deprecate current static factory.

client/src/test/java/com/sinch/sdk/domains/voice/models/requests/CalloutRequestParametersCustomTest.java
JPPortier: I've asked myself same question and it was my 1st implementation ..... :-)
(For Conversation API there is also same kind of extra level depth with only a single field)
For consistency I have kept current builder and extra level to prevent future potential new fields
We can imagine having to function with different parameter signature to provide a "quick access" to the setting in the future; in this case, always for consistency purpose across domains, it should be addressed at "larger scale"

openapi-contracts/src/test/java/com/sinch/sdk/domains/voice/models/dto/svaml/ActionRunMenuDtoTest.java
JPPortier: Yes, fFor this test, only providing a dtmf value with authorized characters was the only purpose

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/SAVMLActionDtoConverter.java
JPPortier: Yes here the getter is the client facing function name and public but the setter is for the dto/generated and not public.
`setAmd` is coming from OAS file defining field `amd`; then generator created the `setAmd` setter
For client facing I choose a more informative naming `AnsweringMachineDetectionEnabled`
1. Name is wrong; it should be `AnsweringMachineDetection` without `Enabled`suffix
2. Class is not consistent with itself: other field `dtmf`  do not have more informative name. It should have been getDualToneMultiFrequency/setDualToneMultiFrequency
Working onto it

client/src/main/com/sinch/sdk/domains/voice/models/svaml/ActionConnectSip.java
JPPortier: This is coming from initial conferenceCalloutRequest schema implementation. OAS is defining a `mohClass` so enums was named at this time. Other schemas using moh are using just a "moh"
You're right naming is not good; and will rename to a more auto documented `MusicOnHoldType` class name

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/SAVMLActionDtoConverter.java
JPPortier: Fixed

client/src/main/com/sinch/sdk/domains/voice/models/svaml/InstructionSetCookie.java
JPPortier: Right 👍 
Fixed (but, by purpose,  let possibility to end user setting a cookie named 'null' ....)

client/src/main/com/sinch/sdk/domains/voice/models/svaml/ActionConnectSip.java
JPPortier: Fixed

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/SAVMLActionDtoConverter.java
JPPortier: Fixed, added: 
- a guard against null pointer onto format
- a validation against null value onto menu constructor because of it does have not a lot of sense to set action to null

client/src/main/com/sinch/sdk/domains/voice/adapters/converters/SAVMLActionDtoConverter.java
asein-sinch: Thanks for the clarification client vs DTO!

asein-sinch (approved):

PR #44, opened 2024-01-25
feature: Remove restriction onto unified credentials from SinchClient…

JPPortier:
… and delegate validation at domain level

### DIFF
--- a/client/src/main/com/sinch/sdk/SinchClient.java
+++ b/client/src/main/com/sinch/sdk/SinchClient.java
@@ -143,9 +143,6 @@ public class SinchClient {
   }
 
   private void checkConfiguration(Configuration configuration) throws NullPointerException {
-    Objects.requireNonNull(configuration.getKeyId(), "'keyId' cannot be null");
-    Objects.requireNonNull(configuration.getKeySecret(), "'keySecret' cannot be null");
-    Objects.requireNonNull(configuration.getProjectId(), "'projectId' cannot be null");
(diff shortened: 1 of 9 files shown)


Dovchik (commented):
Can you add a descriptions about what drives this change?

asein-sinch (approved):
LGTM

JPPortier:
> Can you add a descriptions about what drives this change?

Sure.

### About not checking unified credentials at SinchClient level

When using only a single domain not based onto unified credentials: initial implementation require to set faked values for project (keyId, keySecret and projectId) just to pass the check against null value. This was causing something like
```
    SinchClient client = new SinchClient(Configuration.builder()
                                    .setKeyId("unused but required value")
                                    .setKeySecret("unused but required value")
                                    .setProjectId("unused but required value")
                                    .setApplicationKey("my true application key because I want to use Verification service")
                                    .setApplicationSecret("my true application secret because I want to use Verification service")
                                    .build());
```
But because of client instance is for Verification/Voice only:
```
    SinchClient client = new SinchClient(Configuration.builder()
                                    .setApplicationKey("my true application key because I want to use Verification service")
                                    .setApplicationSecret("my true application secret because I want to use Verification service")
                                    .build());
```
is enough.

### About moving validation at domain level

Finally because previous validation is removed and to keep a validation about consistency onto init; the validation is moved to domain responsibility. This is having sense: `Who know best that the domain itself it requirements if it is no longer a common pattern with unified credentials ?`

Dovchik (approved):

PR #50, opened 2024-02-06
Verification API: Interactive sample, javadoc enhancements and code cleaning

JPPortier:


client/src/main/com/sinch/sdk/domains/verification/adapters/StatusService.java
```diff
@@ -30,16 +28,10 @@ protected QueryVerificationsApi getApi() {
...
-      throw new ApiException("Unexpected entity: " + identity);
-    }
-    NumberIdentity numberIdentity = (NumberIdentity) identity;
+  public VerificationReport get(NumberIdentity identity, VerificationMethodType method) {
 
     return VerificationsDtoConverter.convert(
-        getApi()
-            .verificationStatusByIdentity(
-                numberIdentity.getType(), numberIdentity.getEndpoint(), method.value()));
+        getApi().verificationStatusByIdentity("number", identity.getEndpoint(), method.value()));
```
asein-sinch, line 34: I would prefer to see identity.getType() or NumberIdentity.TYPE instead of hardcoding `number` here.

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationsService.java
```diff
@@ -38,18 +36,12 @@ public StartVerificationResponse start(StartVerificationRequestParameters parame
...
-    NumberIdentity numberIdentity = (NumberIdentity) identity;
+      NumberIdentity identity, VerificationReportRequestParameters parameters) {
 
     return VerificationsDtoConverter.convert(
         getApi()
             .reportVerificationByIdentity(
-                numberIdentity.getType(),
-                numberIdentity.getEndpoint(),
-                VerificationsDtoConverter.convert(parameters)));
+                "number", identity.getEndpoint(), VerificationsDtoConverter.convert(parameters)));
```
asein-sinch, line 44: Same as above: I would prefer to see identity.getType() or NumberIdentity.TYPE instead of hardcoding `number` here. The identity type should not be decorrelated from the parameter.

client/src/main/com/sinch/sdk/domains/verification/models/NumberIdentity.java
```diff
@@ -1,18 +1,27 @@
...
   public String getEndpoint() {
     return endpoint;
   }
 
   /**
    * @param endpoint An E.164-compatible phone number.
    */
-  public NumberIdentity(String endpoint) {
+  private NumberIdentity(String endpoint) {
     super("number");
```
asein-sinch, line 25: This is not part of this PR and it's over-engineered , but instead of using a String here lost in the middle of the class, can we have a psfs instead?

openapi-contracts/src/test/resources/domains/verification/v1/error-response.json
```diff
@@ -0,0 +1,5 @@
+{
+  "errorCode": 40400,
```
asein-sinch, line 2: That's a strange code 😅

client/src/test/java/com/sinch/sdk/models/E164PhoneNumberTest.java
```diff
@@ -30,12 +30,12 @@ class E164PhoneNumberTest {
...
   @Test
   void stringValue() {
-    assertEquals(new E164PhoneNumber(phoneNumber).stringValue(), phoneNumber);
+    assertEquals(E164PhoneNumber.valueOf(phoneNumber).stringValue(), phoneNumber);
   }
 
   @Test
   void valueOf() {
-    assertEquals(E164PhoneNumber.valueOf(phoneNumber), new E164PhoneNumber(phoneNumber));
+    assertEquals(E164PhoneNumber.valueOf(phoneNumber), E164PhoneNumber.valueOf(phoneNumber));
```
asein-sinch: I bet it works!

sample-app/src/main/java/com/sinch/sample/verification/VerificationsSampleFlowInteractive.java
```diff
@@ -0,0 +1,162 @@
...
+   *
+   * @param service Verification service
+   * @param phoneNumber Destination phone number
+   * @return Verification ID
+   */
+  private VerificationId startSmsVerification(
+      VerificationsService service, E164PhoneNumber phoneNumber) {
+
+    var parameters =
+        StartVerificationRequestParameters.builder(VerificationMethodType.SMS)
```
asein-sinch: I find it strange the flashCall has its own request parameters class because it has one extra parameter, but for the other ones, the method type needs to be specified.
What would happen if someone does `StartVerificationRequestParameters.builder(VerificationMethodType.FLASH_CALL)`? I don't think they will have the possibility to set a dial timeout whereas the code looks legit.

sample-app/src/main/java/com/sinch/sample/verification/VerificationsSampleFlowInteractive.java
asein-sinch, line 1: Super nice example! 🌟

client/src/main/com/sinch/sdk/domains/verification/adapters/StatusService.java
JPPortier: Yes by design it was removed by current PR.
The `numberIdentity.getEndpoint()` do not have to be public; this is an implementation detail because of part of the URL to be addressed by service.
From SDK's user perspective it does not have sense to having access to it and could cause misunderstanding or worst could be used because of public now and will cause issue with SDK usage if change in the future.
Even the `NumberIdentity.TYPE` will have to be public and could cause same kind of issue later.
The only way to avoid this being public but not hardcoded will come when design started with Conversation will spread onto this service: `models` as `interfaces` and implementation onto adapters side with known-only-by-service-implementation contracts

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationsService.java
JPPortier: see previous comment about "number" hard coded string usage and why not making string shared/public

client/src/main/com/sinch/sdk/domains/verification/models/NumberIdentity.java
JPPortier: see previous comment about "number" hard coded string usage and why not making string shared/public

client/src/test/java/com/sinch/sdk/models/E164PhoneNumberTest.java
JPPortier: 😕 
Fixed

sample-app/src/main/java/com/sinch/sample/verification/VerificationsSampleFlowInteractive.java
JPPortier: Yes, and it break pattern in place with other domains.
WIP to fix it

JPPortier: Done

client/src/main/com/sinch/sdk/domains/verification/models/requests/StartVerificationCalloutRequestParameters.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @since 1.0
+ */
+public class StartVerificationCalloutRequestParameters extends StartVerificationRequestParameters {
+
+  private StartVerificationCalloutRequestParameters(
+      OptionalValue<Identity> identity,
+      OptionalValue<VerificationReference> reference,
+      OptionalValue<String> custom) {
+    super(identity, OptionalValue.of(VerificationMethodType.CALLOUT), reference, custom);
```
asein-sinch: Do you need to keep the `method` parameter as an Optional?

sample-app/src/main/java/com/sinch/sample/verification/verifications/Start.java
```diff
@@ -34,10 +35,14 @@ public void run() {
     StartVerificationRequestParameters.Builder<?> builder;
 
     if (method != VerificationMethodType.CALLOUT) {
-      builder = StartVerificationRequestParameters.builder(method);
+      builder = StartVerificationCalloutRequestParameters.builder();
```
asein-sinch, line 40: I'm not sure to understand the logic here: it will always be a `callout` verification either on the line 38, either on line 40. The lines 42 and 44 seem unreachable

client/src/test/java/com/sinch/sdk/models/E164PhoneNumberTest.java
```diff
@@ -35,7 +35,9 @@ void stringValue() {
 
   @Test
   void valueOf() {
-    assertEquals(E164PhoneNumber.valueOf(phoneNumber), E164PhoneNumber.valueOf(phoneNumber));
+    assertEquals(
+        E164PhoneNumber.valueOf(phoneNumber).stringValue(),
+        E164PhoneNumber.valueOf(phoneNumber).stringValue());
```
asein-sinch: Still, I don't understand this test: both terms are the same

client/src/main/com/sinch/sdk/domains/verification/models/NumberIdentity.java
JPPortier: I missed a point about `NumberIdentity` usage and over engineering aspect (if I understood right)
This is just to not having some kind of technical debt in the future: Currently only `number identity` is supported, but OAS claim we could have more: 
```
            "description": "Type of identity, e.g. 'number', 'email' etc.",
```
So endpoint value will be different (and again: it is a detail related to HTTP endpoint not having sense to become public)
This mean for Java SDK:
- having a common identity base class
- defining a dedicated `NumberIdentity`
- when time to support something like `EmailIdentity` will arise: just implement it. 
 
This to not breaking existing implementation dedicated to Number from SDK user perspective (let say thank to Uncle Bob about the `O` from `SOLID`: https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle

client/src/main/com/sinch/sdk/domains/verification/models/requests/StartVerificationCalloutRequestParameters.java
JPPortier: Really not.
Thank you
Fixed

sample-app/src/main/java/com/sinch/sample/verification/verifications/Start.java
JPPortier: Yes.
Sample purpose: just change the `method` assignation from line 33 to change the request to be sent to server
Body of function is for educational and highlight how to handle dedicated initialization.

client/src/test/java/com/sinch/sdk/models/E164PhoneNumberTest.java
JPPortier: Agree; sorry to waste your time
Fixed by removing the test (the `valueOf` is indirectly tested by previous test related to `stringValue`)

client/src/main/com/sinch/sdk/domains/verification/models/requests/StartVerificationRequestParameters.java
```diff
@@ -117,9 +116,9 @@ public static class Builder<B extends Builder<B>> {
      * @param method The type value
      * @since 1.0
      */
-    protected Builder(VerificationMethodType method) {
-      this.method = OptionalValue.of(method);
-    }
+    /*protected Builder(VerificationMethodType method) {
```
asein-sinch: You can remove the commented methods and their bodies when abstract

sample-app/src/main/java/com/sinch/sample/verification/verifications/Start.java
asein-sinch: Even for educational purpose, I don't think this code will reach its purpose: 
 - when executing the test as is, the method is set to `SMS` (line 33). When reaching the `if` on line 37, the method NOT callout so the builder will become a `CALLOUT` method
 - if the user changes the method to `CALLOUT` on line 33, the `if` clause on line 37 will be false and lead to the line 39 which will resolve to true: the builder will become a `CALLOUT` method but by entering in the `SMS` clause 😵‍💫
 - and when changing to `SEAMLESS` or `FLASH_CALL` it will result in the same output as the first bullet

```suggestion
      if (method == VerificationMethodType.CALLOUT) {
        builder = StartVerificationCalloutRequestParameters.builder();
      } else if (method == VerificationMethodType.SMS) {
        builder = StartVerificationSMSRequestParameters.builder();
      } else if (method == VerificationMethodType.SEAMLESS) {
        builder = StartVerificationSeamlessRequestParameters.builder();
      } else if (method == VerificationMethodType.FLASH_CALL) {
        builder = StartVerificationFlashCallRequestParameters.builder();
      }
```

client/src/main/com/sinch/sdk/domains/verification/models/requests/StartVerificationRequestParameters.java
JPPortier: Removed

sample-app/src/main/java/com/sinch/sample/verification/verifications/Start.java
JPPortier: Right, I do not remember where I read/listen this a long time ago: `the "cut and paste" function was the worst invention`
I have to agree 😞

JPPortier: Fixed the fix

asein-sinch: 🙌

asein-sinch (approved):

PR #55, opened 2024-02-14
Conversation: contact

JPPortier:


client/src/main/com/sinch/sdk/domains/conversation/ContactService.java
```diff
@@ -0,0 +1,154 @@
...
+   * <p>Note that, in order to retrieve a WhatsApp display name, you can use the Get a Contact or
+   * List Contacts operations, which will populate the display_name field of each returned contact
+   * with the WhatsApp display name (if the name is already stored on the server and the
+   * display_name field has not been overwritten by the user).
+   *
+   * @param parameters The ID of the contact to be used to check profile information
+   * @return The profile name
+   * @since 1,0
+   */
+  String getChannelProfileByContact(GetChannelProfileByContactRequestParameters parameters);
```
asein-sinch: Looking at the documentation, the parameter is a `contact_id`. Do you think the method can be renamed `getChannelProfileByContactId` for full disclosure ?

client/src/main/com/sinch/sdk/domains/conversation/ContactService.java
```diff
@@ -0,0 +1,154 @@
...
+   * List Contacts operations, which will populate the display_name field of each returned contact
+   * with the WhatsApp display name (if the name is already stored on the server and the
+   * display_name field has not been overwritten by the user).
+   *
+   * @param parameters The identity as specified by the channel to be used to check profile
+   *     information
+   * @return The profile name
+   * @since 1,0
+   */
+  String getChannelProfileByChannel(GetChannelProfileByChannelRequestParameters parameters);
```
asein-sinch: Here to, rather that a channel object, the parameter will be a channel identity (a list of channel identities in fact). Similarly to the method above, do you think we can rename this method `getChannelProfileByChannelIdentity` which would be understood as "get contact's channel profile by its channel identity"

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
```diff
@@ -0,0 +1,159 @@
...
+
+    if (null != client && !supportedChannelForGetProfile.contains(client.getChannel())) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid channel value '%s'. Channel have to be in list '%s'",
+              client.getChannel(), supportedChannelForGetProfile));
+    }
+  }
+
+  private TokenPageNavigator handleEmptyPageToken(TokenPageNavigator token) {
```
asein-sinch: Maybe this method can be moved to a upper level as other APIs may return empty page tokens

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
```diff
@@ -0,0 +1,159 @@
...
+  public Contact update(Contact iclient) {
+    ContactImpl client = (ContactImpl) iclient;
+
+    return ContactDtoConverter.convert(
+        getApi()
+            .contactUpdateContact(
+                configuration.getProjectId(),
+                client.getId(),
+                ContactDtoConverter.convert(client),
+                null));
```
asein-sinch, line 107: So in the end, what is your decision about the updateMask? If you don't want to support it at all, then you may remove it from the signature. Otherwise, you need to change this function call to include the list of properties

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
```diff
@@ -0,0 +1,159 @@
...
+            .contactGetChannelProfile(
+                configuration.getProjectId(), ContactDtoConverter.convert(client)));
+  }
+
+  private void checkGetChannelProfileChannelType(BaseGetChannelProfileRequestParameters client) {
+
+    if (null != client && !supportedChannelForGetProfile.contains(client.getChannel())) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid channel value '%s'. Channel have to be in list '%s'",
```
asein-sinch: Typo: `Channel has to be in list`

client/src/main/com/sinch/sdk/domains/conversation/adapters/models/requests/ContactCreateRequestParametersImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import com.sinch.sdk.domains.conversation.adapters.models.ChannelIdentityImpl;
+import com.sinch.sdk.domains.conversation.models.ChannelType;
+import com.sinch.sdk.domains.conversation.models.LanguageType;
+import com.sinch.sdk.domains.conversation.models.requests.ContactCreateRequestParameters;
+import java.util.Collection;
+
+public class ContactCreateRequestParametersImpl extends ContactCreateUpdateRequestParametersImpl
+    implements ContactCreateRequestParameters {
+
+  protected ContactCreateRequestParametersImpl(
```
asein-sinch, line 14: At least the channel_identities and the language are required

client/src/main/com/sinch/sdk/domains/conversation/adapters/models/responses/AppImpl.java
asein-sinch, line 1: Is it linked to this PR about contact?

sample-app/src/main/java/com/sinch/sample/conversation/contact/GetChannelProfile.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  public void run() {
+
+    var parametersByContact =
+        GetChannelProfileByContactRequestParameters.builder()
+            .setAppId(conversationAppId)
+            .setChannel(ChannelType.MESSENGER)
+            .setContactId("33628254417")
```
asein-sinch: Just FYI, the contact ID is not a phone number but an ULID

sample-app/src/main/java/com/sinch/sample/conversation/contact/Update.java
```diff
@@ -0,0 +1,66 @@
...
+    LOGGER.info("Update contact '%s'".formatted(conversationContactId));
+
+    var contact =
+        Contact.builder()
+            .setId(conversationContactId)
+            .setChannelIdentities(
+                Arrays.asList(
+                    ChannelIdentity.builder()
+                        .setChannel(ChannelType.MESSENGER)
+                        .setIdentity("+33123456789")
```
asein-sinch: Should not be a phone number for messenger but the messenger ID (same for instagram, and maybe also for kakao talk TBC)

client/src/main/com/sinch/sdk/domains/conversation/ContactService.java
JPPortier: Right
Done

client/src/main/com/sinch/sdk/domains/conversation/ContactService.java
JPPortier: Right 
Done

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
JPPortier: Of course !
Thank you, done.

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
JPPortier: About updateMask: still no need to be implemented. Without mask, backend will handle only fields present from payload so, mask is not required to update any field.
About function prototype: thisd is the call to "underlying API interface " generated by OAS generator tool; so this file is fully in it "adaptor's role": it hide the mask usage:
-  SDK's user is calling `service.update()` without any mask reference just the `Contact` instance with fields to be updated
- adaptor is calling underlying API with a null mask value

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
JPPortier: Fixed

client/src/main/com/sinch/sdk/domains/conversation/adapters/models/requests/ContactCreateRequestParametersImpl.java
JPPortier: Yes; this is managed during serialization.
Because of Builder pattern usage , required or not parameters won't be highlighted here and delegating required validation at serialization level enable:
- to not duplicate business here in addition to serialization layer
- if field required or not status change, there is no change onto this class: serialization (generated source) will do
- enable to store part of a fields within an object without business logic; Until they are really consumed then BL will be applied (this is not for this dedicated object:; it is a general rule common to this repository)

client/src/main/com/sinch/sdk/domains/conversation/adapters/models/responses/AppImpl.java
JPPortier: Not really sorry. This is a refactoring to align App and Contact interface  + Impl usage
(it come from a dedicated commit : https://github.com/sinch/sinch-sdk-java/pull/55/commits/e8c46cddae126489c3f9cd3a44e0b74026c1dee7)

sample-app/src/main/java/com/sinch/sample/conversation/contact/GetChannelProfile.java
JPPortier: Modified sample

sample-app/src/main/java/com/sinch/sample/conversation/contact/Update.java
JPPortier: Modified sample

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
asein-sinch: Ok, so if I understand well, we can state that the Java SDK is not supporting one of the API feature which is to send an `update_mask` along with the update request body, right?

sample-app/src/main/java/com/sinch/sample/conversation/contact/Update.java
asein-sinch: Maybe your forgot a commit somewhere, I still see the phone number in the example 😞

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
JPPortier: Exact: due to Java compilation and struct this is not required:
- objects are immutable: so you cannot just get an object and edit one a dedicated field to send back with the mas to server
- you have to create a new object with the field you want to edit and the value you want to change: so at this point it replace the requirement to set a mask in addition to set the only required field
- because of SDK is aware to detect which fields was really set onto object instance (thank to OptionalValue dedicated class); it just sent to server fields to be modified

sample-app/src/main/java/com/sinch/sample/conversation/contact/Update.java
JPPortier: 😠 
Done

asein-sinch (approved):

PR #62, opened 2024-02-23
DEVEXP-223: SMS / support service plan ID

JPPortier:
Support service plan ID/token at configuration level
When defined: super-sed unified credentials and switch to service plan usage
The last commit is THE simple feature support

All previous commit are refactoring level and enable to:
- move from away "all in one" at configuration level
- services use dedicated/specialized classes in place of "all in one" configuration

This refactoring will enable incoming task supporting multiple credentials


client/src/main/com/sinch/sdk/SinchClient.java
```diff
@@ -86,57 +92,105 @@ public SinchClient(Configuration configuration) {
             "%s (%s) started with projectId '%s'",
             versionProperties.getProperty(PROJECT_NAME_KEY),
             versionProperties.getProperty(PROJECT_VERSION_KEY),
-            configuration.getProjectId()));
+            this.configuration
```
asein-sinch: I'm wondering why this logger only mentions the projectId as credentials. Moreover, as we have decided to allow to override the credentials as API function level, the projectId value logged here may have no meaning in the execution context.

client/src/main/com/sinch/sdk/models/Configuration.java
```diff
@@ -1,52 +1,37 @@
...
+  private final SmsContext smsContext;
+  private final VerificationContext verificationContext;
+  private final VoiceContext voiceContext;
 
   private Configuration(
-      String keyId,
-      String keySecret,
-      String projectId,
+      UnifiedCredentials unifiedCredentials,
+      ApplicationCredentials applicationCredentials,
```
asein-sinch, line 20: Thinking out loud: here we are considering a user will always use the same application for Verification and Voice API and I'm wondering if there are some cases where customer use a different lifecycle for Verification applications and Voice applications.

client/src/main/com/sinch/sdk/SinchClient.java
```diff
@@ -86,57 +92,105 @@ public SinchClient(Configuration configuration) {
...
   private void handleDefaultSmsSettings(
       Configuration configuration, Properties props, Configuration.Builder builder) {
 
-    if (null == configuration.getSmsUrl() && props.containsKey(SMS_SERVER_KEY)) {
-      builder.setSmsUrl(props.getProperty(SMS_SERVER_KEY));
+    String smsUrl = configuration.getSmsContext().map(SmsContext::getSmsUrl).orElse(null);
+
+    SMSRegion smsRegion = configuration.getSmsContext().map(SmsContext::getSmsRegion).orElse(null);
+
+    // service plan ID activated: use dedicate server
```
asein-sinch: Typo: `dedicated`

client/src/main/com/sinch/sdk/auth/adapters/BearerAuthManager.java
```diff
@@ -1,115 +1,39 @@
...
-    authManagers =
-        Stream.of(new AbstractMap.SimpleEntry<>(SCHEMA_KEYWORD_BASIC, basicAuthManager))
-            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+  public BearerAuthManager(String token) {
+    this.token = token;
   }
 
   public String getSchema() {
-    return SCHEMA_KEYWORD_BEARER;
+    return SCHEMA_KEYWORD_BASIC;
```
asein-sinch: Why this change? Is it a hack to say this is a static token, versus the OAuth2 authentication where the bearer token needs to be refreshed ?

client/src/main/com/sinch/sdk/domains/numbers/adapters/NumbersService.java
```diff
@@ -23,15 +28,24 @@ public class NumbersService implements com.sinch.sdk.domains.numbers.NumbersServ
...
 
-    this.configuration = configuration;
+    LOGGER.fine("Activate numbers API with server='" + context.getNumbersServer().getUrl() + "'");
+
+    this.credentials = credentials;
+    this.context = context;
     this.httpClient = httpClient;
-    AuthManager basicAuthManager = new BasicAuthManager(configuration);
+
+    AuthManager basicAuthManager =
```
asein-sinch, line 47: I missed this one in the previous PRs: are you using only basic authentication for the Numbers API, not OAuth2 ?

client/src/main/com/sinch/sdk/domains/sms/adapters/SMSService.java
```diff
@@ -1,24 +1,33 @@
...
 import java.util.stream.Stream;
 
 public class SMSService implements com.sinch.sdk.domains.sms.SMSService {
-  private static final String SECURITY_SCHEME_KEYWORD_SMS = "BearerAuth";
 
-  private final Configuration configuration;
+  private static final Logger LOGGER = Logger.getLogger(SMSService.class.getName());
+
+  private static final String SECURITY_SCHEME_KEYWORD_SMS = "BearerAuth";
+  private final String uriPathID;
```
asein-sinch: Naming is complicated! I had to go to this class to understand what a `uriPathID` was.
Small remark: it may be renamed to `uriPathId` to be consistent with the other `projectId` and `servicePlanId`

Suggestion: The pattern in the Services is different from numbers. When you switch to auto-generated code, this will be problematic. I would suggest you stick with one of the two implementations
 - the numbers one: you build a UnifiedCredentials object with only the projectId filled with the servicePlanId value, so that the services will fill the url with `credentials.getProjectId()`
  - the SMS one: you extract the "uriPathId" which here can be either the projectId either the servicePlanId and you give this value as parameters to the services instead of a UnifiedCredentials object since the services only use the projectId value from it

client/src/test/java/com/sinch/sdk/domains/voice/adapters/VoiceServiceTest.java
```diff
@@ -4,34 +4,66 @@
...
+            .setApplicationKey(null)
+            .setApplicationSecret("foo secret")
+            .build();
+    VoiceContext context = VoiceContext.builder().build();
+
     Exception exception =
         assertThrows(
             IllegalArgumentException.class,
-            () -> new VerificationService(configuration, httpClient));
+            () -> new VoiceService(credentials, context, httpClient));
```
asein-sinch, line 48: Better 😅

client/src/main/com/sinch/sdk/SinchClient.java
JPPortier: Right 👍 
Removed

client/src/main/com/sinch/sdk/models/Configuration.java
JPPortier: Exact; technically user could choose different application for each domain and with current implementation it is not possible with same Sinch client

client/src/main/com/sinch/sdk/SinchClient.java
JPPortier: fixed

client/src/main/com/sinch/sdk/auth/adapters/BearerAuthManager.java
JPPortier: No hack here ... you pointed an error 👍 
Fixed

client/src/main/com/sinch/sdk/domains/numbers/adapters/NumbersService.java
JPPortier: Right, due to OAS specs used to generate API calls
```
security:
  - BasicAuth: []
 ```
Need to enhance OAS specs and SDK (ticket created)

client/src/main/com/sinch/sdk/domains/sms/adapters/SMSService.java
JPPortier: Refactored with `uriUUID` name and spread across all services to replace `credentials.getProjectID()` usage

client/src/main/com/sinch/sdk/domains/numbers/adapters/NumbersService.java
asein-sinch: Ok 😞

client/src/main/com/sinch/sdk/domains/sms/adapters/SMSService.java
asein-sinch: Or... `uuidPathParameter` 😁

asein-sinch (approved):
Nice work supporting the ServicePlanId and I really like the "context" to manage the URLs and regions 👏

PR #63, opened 2024-02-27
feature: Remove 'Configuration' usage and move conversation to dedicated 'ConversationContext'

JPPortier:
- rebase conversation onto main
- using `context` in place of `configuration`

### DIFF
--- a/client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
+++ b/client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
@@ -33,7 +33,7 @@ import java.util.Map;
 
 public class ContactService implements com.sinch.sdk.domains.conversation.ContactService {
 
-  private final String uriPathID;
+  private final String uriUUID;
   private final ContactApi api;
 
   private final Collection<ChannelType> supportedChannelForGetProfile =
@@ -41,11 +41,11 @@ public class ContactService implements com.sinch.sdk.domains.conversation.Contac
           ChannelType.INSTAGRAM, ChannelType.MESSENGER, ChannelType.VIBER, ChannelType.LINE);
 
   public ContactService(
-      String uriPathID,
+      String uriUUID,
       ConversationContext context,
       HttpClient httpClient,
       Map<String, AuthManager> authManagers) {
-    this.uriPathID = uriPathID;
+    this.uriUUID = uriUUID;
     this.api = new ContactApi(httpClient, context.getServer(), authManagers, new HttpMapper());
   }
 
@@ -68,7 +68,7 @@ public class ContactService implements com.sinch.sdk.domains.conversation.Contac
     String pageToken = client.getPageToken().orElse(null);
 
     ListContactsResponseDto response =
-        getApi().contactListContacts(uriPathID, pageSize, pageToken, externalId, channel, identity);
+        getApi().contactListContacts(uriUUID, pageSize, pageToken, externalId, channel, identity);
 
     Pair<Collection<Contact>, TokenPageNavigator> content = ContactDtoConverter.convert(response);
 
@@ -76,19 +76,19 @@ public class ContactService implements com.sinch.sdk.domains.conversation.Contac
   }
 
   public Contact get(String contactId) {
-    return ContactDtoConverter.convert(getApi().contactGetContact(uriPathID, contactId));
+    return ContactDtoConverter.convert(getApi().contactGetContact(uriUUID, contactId));
   }
 
   public Contact create(ContactCreateRequestParameters iclient) {
     ContactCreateRequestParametersImpl client = (ContactCreateRequestParametersImpl) iclient;
 
     return ContactDtoConverter.convert(
-        getApi().contactCreateContact(uriPathID, ContactDtoConverter.convertForCreate(client)));
+        getApi().contactCreateContact(uriUUID, ContactDtoConverter.convertForCreate(client)));
   }
 
   public void delete(String contactId) {
 
-    getApi().contactDeleteContact(uriPathID, contactId);
+    getApi().contactDeleteContact(uriUUID, contactId);
   }
 
   public Contact update(Contact iclient) {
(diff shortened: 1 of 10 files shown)


client/src/test/java/com/sinch/sdk/domains/conversation/adapters/AppServiceTest.java
```diff
@@ -25,24 +25,24 @@
...
 public class AppServiceTest extends BaseTest {
 
+  @Mock ConversationContext context;
   @Mock AppApi api;
-  Configuration configuration = Configuration.builder().setProjectId("").build();
   @Mock HttpClient httpClient;
   @Mock Map<String, AuthManager> authManagers;
 
   AppService service;
+  String uriPartID = "foovalue";
```
asein-sinch, line 34: What a fancy name 😁

client/src/test/java/com/sinch/sdk/models/ConfigurationBuilderTest.java
```diff
@@ -23,6 +26,12 @@ void build() {
             .setNumbersContext(NumbersContext.builder().setNumbersUrl(NUMBERS_SERVER).build())
             .setSmsContext(
                 SmsContext.builder().setSmsRegion(SMS_REGION).setSmsUrl(SMS_SERVER).build())
+            .setConversationContext(
+                ConversationContext.builder()
+                    .setRegion(CONVERSATION_REGION)
```
asein-sinch, line 31: As the region is an optional parameter, do you think it's worth to add a test to verify the default value when not specified ?

client/src/main/com/sinch/sdk/domains/conversation/adapters/ConversationService.java
```diff
@@ -15,45 +19,55 @@ public class ConversationService implements com.sinch.sdk.domains.conversation.C
...
+      HttpClient httpClient) {
+
+    Objects.requireNonNull(credentials, "Credentials must be defined");
+    Objects.requireNonNull(context, "Context must be defined");
+    StringUtil.requireNonEmpty(credentials.getKeyId(), "'keyId' must be defined");
+    StringUtil.requireNonEmpty(credentials.getKeySecret(), "'keySecret' must be defined");
+    StringUtil.requireNonEmpty(credentials.getProjectId(), "'projectId' must be defined");
+    StringUtil.requireNonEmpty(context.getUrl(), "'url' must be defined");
+    StringUtil.requireNonEmpty(
+        context.getTemplateManagementUrl(), "'templateManagementUrl' must be defined");
```
asein-sinch, line 42: What happens in the case a user is not willing to use the message templates?

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
```diff
@@ -26,26 +26,27 @@
...
-import com.sinch.sdk.models.Configuration;
+import com.sinch.sdk.models.ConversationContext;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 
 public class ContactService implements com.sinch.sdk.domains.conversation.ContactService {
 
-  private final Configuration configuration;
+  private final String uriPathID;
```
asein-sinch: The same variable is named `uriUUID` on the main branch for other APIs. Which one will win?

asein-sinch (approved):

client/src/test/java/com/sinch/sdk/models/ConfigurationBuilderTest.java
JPPortier: This test already exist (but not in the scope of the Builder: default value is managed by context itself):
- context is having a default value if not set: https://github.com/sinch/sinch-sdk-java/blob/feat/conversation-rebase-main/client/src/main/com/sinch/sdk/models/SmsContext.java#L13
- test against default value: https://github.com/sinch/sinch-sdk-java/blob/feat/conversation-rebase-main/client/src/test/java/com/sinch/sdk/models/SmsContextTest.java#L14

client/src/main/com/sinch/sdk/domains/conversation/adapters/ConversationService.java
JPPortier: Currently by using the SinchClient  there is always a value set: https://github.com/sinch/sinch-sdk-java/blob/feat/conversation-rebase-main/client/src/main/com/sinch/sdk/SinchClient.java#L235
But you're right: This will have to refactored when we will have to support context at request level

client/src/main/com/sinch/sdk/domains/conversation/adapters/ContactService.java
JPPortier: 😕 both exists (for SMS)...
refactored to use `uriUUID` for Conversation (not refactored for SMS: will have to be adressed outside of this feature dedicated branch)

client/src/test/java/com/sinch/sdk/models/ConfigurationBuilderTest.java
asein-sinch: But here I was talking about the Conversation API, not the SMS. I'll add a new comment at the right place

client/src/main/com/sinch/sdk/models/ConversationContext.java
```diff
@@ -0,0 +1,161 @@
...
+
+/** Execution context related to Conversation domains */
+public class ConversationContext {
+
+  private final ConversationRegion region;
+  private final String url;
+  private final String templateManagementUrl;
+
+  private ConversationContext(ConversationRegion region, String url, String templateManagementUrl) {
+    this.region = region;
```
asein-sinch, line 13: Contrary to SMS, this constructor is private, but I think the region should not be null. Apart if you say it's defined in a configuration file and the value will never be null.

JPPortier: 1. SMS constructor should be private (will create a dedicated PR related to SMS to avoid in this feature branch only)
2. A default value is currently set by config file (https://github.com/sinch/sinch-sdk-java/blob/feat/conversation-rebase-main/client/resources/config-default.properties#L21)

Ideally SMS should not set a default value from constructor: with future feature (overriding request options at endpoint call level),  this will prevent to have some kind of default config (with region set) and only some context properties defined (not region)

asein-sinch (approved):

PR #64, opened 2024-02-28
SMS refactoring

JPPortier:


client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -34,6 +34,15 @@ void defaultSmsUrlAvailable() {
     assertNotNull(client.getConfiguration().getSmsContext().get().getSmsUrl());
   }
 
+  @Test
+  void defaultSmsRegion() {
+    Configuration configuration =
+        Configuration.builder().setKeyId("foo").setKeySecret("foo").setProjectId("foo").build();
+    SinchClient client = new SinchClient(configuration);
+    System.out.println(client.getConfiguration().getSmsContext().get().getSmsRegion());
+    assertNotNull(client.getConfiguration().getSmsContext().get().getSmsRegion());
```
asein-sinch: Is it safe to assume the not null value is 'us' ?

asein-sinch (dismissed):

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: You're right: checking now against `SMSRegion.US` (and added a test about Voice default region too)

asein-sinch (approved):

PR #66, opened 2024-03-12
refactor: Rename 'VerificationStatusService.get' functions to dedicat…

JPPortier:
…ed function names


asein-sinch (approved):
Although Java supports overloading, I agree it's best to have different method names since the API endpoints are on different paths: it makes a clear mapping with the developer documentation 👍

PR #72, opened 2024-03-28
Feat/conversation messages

JPPortier:
Synch with latest 
- OAS file
- latest code generator tool

This PR is sync point for feature branch
It just provide an intermediate state because of:
- need to follow latest OAS and tool states
- generate classes do not provide a lot of added value over simple JSON payload usage: it require too much from SDK's user  to take care about JSON structure in place of dedicated classes auto-documenting content
- provide a starting point for Messages with unit tests to ensure regression with further features to be implemented

PR #74, opened 2024-04-19
refactoring checkpoint: Internal vs public oneOf integration

JPPortier:
big big refactoring to improve DX usage and reduce layers to fill for payloads
Big picture:
- Ingest latest spec file version
- Reduce objects complexity creation : e.g [ChoiceItem or ProductItem creation improvement](https://github.com/sinch/sinch-sdk-java/commit/0e4e61750d06dd4b5bbec6ec885488994a17b627#diff-a161b00d472eac555c48b1ac77ed884abe23c9fd362ca49cf3fdec80c203ae85) from unit test.
- dedicated functions to SendMessage (sendCard, sendText, ...)
- Cannot mix Choice/Product within a list ([comment from previous PR](https://github.com/sinch/sinch-sdk-java/pull/72#discussion_r1544665568))
- `item` layer removed ([comment from previous PR](https://github.com/sinch/sinch-sdk-java/pull/72#discussion_r1544660789))
- `card_message` layer removed for Carousel ([comment from previous PR](https://github.com/sinch/sinch-sdk-java/pull/72#discussion_r1544673685)) 
- `message` layer removed for Carousel ([comment from previous PR](https://github.com/sinch/sinch-sdk-java/pull/72#discussion_r1544752040))

Details: at each dedicated commit level


client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
```diff
@@ -34,7 +43,23 @@ public interface MessagesService {
...
    * @since 1.0
    */
-  SendMessageResponse send(SendMessageRequest request);
+  SendMessageResponse sendCardMessage(SendMessageRequest<CardMessage> request);
+
+  SendMessageResponse sendCarouselMessage(SendMessageRequest<CarouselMessage> request);
+
+  SendMessageResponse sendChoiceMessage(SendMessageRequest<ChoiceMessage> request);
+
+  SendMessageResponse sendCContactInfoMessage(SendMessageRequest<ContactInfoMessage> request);
```
asein-sinch: Typo

client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
asein-sinch, line 1: Nice DX improvement to allow only a single type os message per request 👍

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/WhatsAppInteractiveHeaderMediaBase.java
```diff
@@ -0,0 +1,12 @@
+package com.sinch.sdk.domains.conversation.models.v1.messages;
+
+public interface WhatsAppInteractiveHeaderMediaBase {
+
+  WhatsAppInteractiveHeaderMedia getMedia();
+
+  /** Dedicated Builder */
+  interface Builder<B extends Builder<B>> {
+
+    B setMedia(WhatsAppInteractiveHeaderMedia text);
```
asein-sinch: The variable can be renamed `media` for consistency

openapi-contracts/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/internal/WhatsAppInteractiveDocumentHeaderInternal.java
```diff
@@ -0,0 +1,101 @@
...
+import com.sinch.sdk.domains.conversation.models.v1.messages.WhatsAppInteractiveHeaderMedia;
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+/** Header of the interactive message with document. */
+@JsonDeserialize(builder = WhatsAppInteractiveDocumentHeaderInternalImpl.Builder.class)
+public interface WhatsAppInteractiveDocumentHeaderInternal {
+
+  /** Gets or Sets type */
+  public class TypeEnum extends EnumDynamic<String, TypeEnum> {
```
asein-sinch, line 27: Here the `type` property can be considered as a discriminator and the value should be unique. Does it make sense to create a `EnumDynamic` ?

client/src/main/com/sinch/sdk/domains/conversation/models/v1/contact/request/ContactGetChannelProfileByChannelIdentityRequest.java
```diff
@@ -14,10 +13,7 @@ public class ContactGetChannelProfileByChannelIdentityRequest extends GetChannel
       List<ChannelRecipientIdentity> identities) {
     super(
         OptionalValue.of(appId),
-        OptionalValue.of(
-            Recipient.builder()
-                .setIdentifiedBy(IdentifiedBy.builder().setChannelIdentities(identities).build())
-                .build()),
+        OptionalValue.of(IdentifiedBy.builder().setChannelIdentities(identities).build()),
```
asein-sinch, line 16: Yes, it becomes more readable thanks to the dedicated mapper! 🙌

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/AppMessageMessage.java
```diff
@@ -0,0 +1,3 @@
+package com.sinch.sdk.domains.conversation.models.v1.messages;
+
+public interface AppMessageMessage {}
```
asein-sinch, line 3: I see this interface, but I can't find the similar setup for `ContactMessage`. Do you confirm it will be considered later when you will work on the `ConversationService` (not the one that already exits but a new one that handles the operation related to the Conversation subdomain - createConversation / updateConversation / injectMessage ...)

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/ChoiceItemMapper.java
asein-sinch, line 1: I'm surprised to see a mapper here. Shouldn't it be located in the adapter folder?

client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
```diff
@@ -34,7 +43,23 @@ public interface MessagesService {
    *     com.sinch.sdk.domains.conversation.models.v1.messages.response.SendMessageResponse})
    * @since 1.0
    */
-  SendMessageResponse send(SendMessageRequest request);
```
asein-sinch, line 37: I've found that having a generic `send` message may prove useful. Do you have a reason for deleting it?

openapi-contracts/src/test/resources/domains/conversation/v1/messages/ProductItemDto.json
```diff
@@ -0,0 +1,7 @@
+{
+  "id": "product ID value",
+  "marketplace": "marketplace value",
+  "quantity": 4,
+  "item_price": 3.14159,
```
asein-sinch, line 5: That's an approximation!

openapi-contracts/src/test/resources/domains/conversation/v1/messages/request/AppMessageCarouselRequestDto.json
```diff
@@ -0,0 +1,158 @@
...
+            "url_message": {
+              "title": "title value",
+              "url": "an url value"
+            },
+            "postback_data": "postback_data url"
+          }
+        ]
+      }
+    ],
+    "choices": [
```
asein-sinch, line 46: Tested on messager: although the specification allows it, the server will tell you that only a text message will be allowed
```
The underlying channel reported: Invalid choices in field [carousel_message.choices] (only text choices are allowed)
```

client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
JPPortier: Fixed

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/WhatsAppInteractiveHeaderMediaBase.java
JPPortier: Fixed

openapi-contracts/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/internal/WhatsAppInteractiveDocumentHeaderInternal.java
JPPortier: Yes .....but 🤔 
From final generated source it is a constant and should only be a simple String

Current generated output is due to to OAS description:
- the type is used as the discriminator
- the discriminator is using an enum value but with a single value defined (required to have proper on line doc)

`mustache` templates are generic to handle `enums` and are generating `EnumDynamic`

I created a ticket (DEVEXP-414) to research what could be done during sources generation:
- detect single enum value usage ? not good enough: we could have single value enum def outside of oneOf discriminator 
- detect oneOf + discriminator + single value
- extend generator tool at configuration level to tag field as a constant and not a dynamic enum

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/AppMessageMessage.java
JPPortier: Exact.
Purpose of this PR is to have a checkpoint for SendMessage (lot of OAS sync and generator tool improvement was required)
Next steps are related to the response (yes this PR only manage the request ...) and then at some point will come the Contact service

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/ChoiceItemMapper.java
JPPortier: Yes, thank you to highlight 😄 
- see here: https://github.com/sinch/sinch-sdk-java/pull/74/files/ce69ef98ee9119ceaf3d8d0766924f4c1c9c3387#diff-817a2cf0887484de6b662318ff99b1a6c182bcd0289b17ab807ba715771bc4e3R18
- and here: https://github.com/sinch/sinch-sdk-java/pull/74/files/ce69ef98ee9119ceaf3d8d0766924f4c1c9c3387#diff-a941781d021757a34f695cc3521439d60399476a234da6848cf02a4122e427dcR11

It is by purpose and due to Java language feature
Ready to onboard ?
1. Design choice related to Builder is to have them `protected` so no access outside of package: https://github.com/sinch/sinch-sdk-java/pull/74/files/ce69ef98ee9119ceaf3d8d0766924f4c1c9c3387#diff-4034c204e4044c1db0a7516eccd182da4a0222a6221993e666019e2a95977649R142
The Builder is not `public` so `protected` by default
2. Then the only way to create a Builder is to use the the interface which is public: https://github.com/sinch/sinch-sdk-java/pull/74/files/ce69ef98ee9119ceaf3d8d0766924f4c1c9c3387#diff-0de381f465112f6faca0d0f468521872f339e4cf8b784710c2cfb0f8c9d9f506R56
3. But because we are looking to use mixin to override serialize/unserialize enabling to add the wrapper field for media (`media` JSON payload parts) we need to override the Builder class which is not public....
4. The trick: to avoid breaking the design about public/protected (for all generated sources) ...: create a bridge (this file) within same package (source is not present within openapi-contracts directory, but package name here is identical to `com.sinch.sdk.domains.conversation.models.v1.messages` like https://github.com/sinch/sinch-sdk-java/blob/ce69ef98ee9119ceaf3d8d0766924f4c1c9c3387/openapi-contracts/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/ChoiceItemImpl.java#L1

So what was done: we created a source from directory but at compilation time we are "injecting" it within same the Builder's package: then it have access to protected members and we can override the Builder. 

Why this trick and not having store the file openapi-contracts directory ?
Because of openapi-contracts is a directory where there is only generated source files enabling to remove them easily from one OAS file version to another

A way to avoid this trick could have been to use dynamic loading feature from Java with ServiceLoader and modules description but outside of this task scope

client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
JPPortier: Yes, I did not yet identified utility.
So instead of having to, maintain a feature not yet identified as useful/required I did not retain it.
But I'm interesting if you already identified a feature and no problem to make it public and maintain it.

openapi-contracts/src/test/resources/domains/conversation/v1/messages/ProductItemDto.json
JPPortier: More or less 😉

openapi-contracts/src/test/resources/domains/conversation/v1/messages/request/AppMessageCarouselRequestDto.json
JPPortier: Will check with conversation team, thank you for highlighting (can we imagine that... e.g. kakao will accept a mix ?)

Current json file purpose is to ensure 100% coverage about carousel content from unit tests; so test follow specs but no problem to align in case of feedback are highlighting a specs issue

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/ChoiceItemMapper.java
JPPortier: Another point: I did not choose such trick without 100% unit tests coverage to ensure the mapper and overriding feature are still running across sources lifecycle

client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
asein-sinch: Sure, my use case is the following: when handling the conversation event, I have a bot that reads the input message from the contact. The message that the chatbot will send back is delegated to another service that can produce a TextMessage, a LocationMessage... or anything else. Thus, I need a `send` method that takes any kind of message as a parameter to reply to the contact.

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/ChoiceItemMapper.java
asein-sinch: Thanks you very much for explaining in details, I was so much confused by the usage of the mixin!

asein-sinch (approved):
Great work, we are getting closer!

client/src/main/com/sinch/sdk/domains/conversation/MessagesService.java
JPPortier: Good point.
I think I'll need this kind of requirements too when it will time about webhook supports 
Thank you

PR #78, opened 2024-04-25
Synchronize generated sources with latest OAS

JPPortier:
No new feature; just synchronization point for feature branch


openapi-contracts/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/internal/ListItemOneOfProductInternalImpl.java
```diff
@@ -32,7 +32,7 @@ public ProductItem getProduct() {
   }
 
   @JsonProperty(JSON_PROPERTY_PRODUCT)
-  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
+  @JsonInclude(value = JsonInclude.Include.ALWAYS)
```
asein-sinch, line 35: Shouldn't it also be updated for Choice?

openapi-contracts/src/main/com/sinch/sdk/domains/conversation/models/v1/response/ConversationMessage.java
asein-sinch, line 1: I see that you are moving the contactMessage and the appMessage one level down from the InjectMessageRequest. But as it is a oneOf (the injected message is either from a contact either from an app), is there a way to prevent setting the 2 of them? Or maybe it's not related to the code here as we are in the generated part?

asein-sinch (approved):

openapi-contracts/src/main/com/sinch/sdk/domains/conversation/models/v1/response/ConversationMessage.java
JPPortier: Yes, with this commit it is just synch out of the box from OAS file and generator tool
Manual "curation" about exact location for proper DX is not yet addressed but will be when implementing it

PR #80, opened 2024-05-16
DEVEXP-307: simplify identifiedBy usage

JPPortier:
Simplify interface to fill a recipient when sending a message (in channel identities context : https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/Messages/#tag/Messages/operation/Messages_SendMessage!path=recipient&t=request)

Old sequence:
```
          .setRecipient(
              IdentifiedBy.builder()
                  .setChannelIdentities(
                      Collections.singletonList(
                          ChannelRecipientIdentity.builder()
                              .setChannel(ConversationChannel.MESSENGER)
                              .setIdentity("an identity")
                              .build()))
                  .build())

```
New sequence:
```
          .setRecipient(
              ChannelRecipientIdentities.of(
                  ChannelRecipientIdentity.builder()
                      .setChannel(ConversationChannel.MESSENGER)
                      .setIdentity("an identity")
                      .build()))
```


sample-app/src/main/java/com/sinch/sample/conversation/messages/Send.java
```diff
@@ -44,7 +46,12 @@ public void run() {
                             .setDisplayName("Agent Name")
                             .build())
                     .build())
-            .setRecipient(ContactId.builder().setContactId(conversationContactId).build())
+            .setRecipient(
+                ChannelRecipientIdentities.of(
```
asein-sinch, line 50: Due to the special treatment for this property, the SDK user will write `ChannelRecipientIdentities.of()` instead of `ChannelRecipientIdentities.builder()` which is the general pattern when building the request parameters. How would a SDK user become aware of this pattern for this specific parameter?

JPPortier: Thank you, I asked myself this question 😄 
This is why the Builder pattern is also available: https://github.com/sinch/sinch-sdk-java/pull/80/files#diff-2e9dd90495dafda82c2c06cebcf3589bc64776421ca434040248230180b31862R22
About being aware; only javadoc will help user to "discover" this helper method or a release note dedicated note 

For information: this is an helper trying to mimic j[ava 9 List.of](https://docs.oracle.com/javase/9/docs/api/java/util/List.html#of-E...-) helper method

asein-sinch (approved):

PR #88, opened 2024-06-18
Verification DX

JPPortier:
take away:
- Naming consistency for Verification (coming from OAS and generator)
- Delegation usage for dedicated classes options(coming from OAS and generator)
- Synchronize with doc terms: 
  - callout -> phone call
  - seamless -> data
- Support verification.api.v1 and verification.models.v1
- Handle deprecation of `acceptLanguage` field for `VerificationRequyestEvent` webhook


pom.xml
```diff
@@ -131,10 +132,14 @@
...
                             **.adapters.converters;
                             **.adapters;
+                            **.internal;
+                            **.v*.**;
                         </excludePackageNames>
                         <sourceFileExcludes>
                             <sourceFileExclude>**/pagination/Page*.java</sourceFileExclude>
                             <sourceFileExclude>**/AbstractOpenApiSchema.java</sourceFileExclude>
+                            <sourceFileExclude>**/*Impl.java</sourceFileExclude>
+
```
asein-sinch: Formating

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/webhooks/VerificationResponseEventDtoTest.java
```diff
@@ -3,49 +3,37 @@
...
+  public static VerificationRequestEventResponsePhoneCall expectedPhoneCallRequestEventResponseDto =
+      VerificationRequestEventResponsePhoneCall.builder()
           .setAction(VerificationEventResponseAction.ALLOW)
-          .setCallout(
-              PhoneCallRequestEventResponsePhoneCallContent.builder()
-                  .setCode("4567")
-                  .setSpeech(PhoneCallSpeech.builder().setLocale("the locale").build())
-                  .build())
+          .setCode("4567")
+          .setSpeech(PhoneCallSpeech.builder().setLocale("the locale").build())
```
asein-sinch, line 24: Can we think about going one step further and do something like:
```java
.setSpeechLocale("the locale")
```
Indeed, the documentation states that "speech' defines various properties but there is only one and its value will always be the same 🙃

client/src/main/com/sinch/sdk/domains/verification/api/v1/adapters/VerificationService.java
```diff
@@ -0,0 +1,150 @@
...
+    LocalLazyInit.init();
+  }
+
+  public VerificationService(
+      ApplicationCredentials credentials, VerificationContext context, HttpClient httpClient) {
+
+    // Currently, we are not supporting unified credentials: ensure application credentials are
+    // defined
+    Objects.requireNonNull(
+        credentials, "Verification service require application credentials to be defined");
```
asein-sinch:
```suggestion
        credentials, "Verification service requires application credentials to be defined");
```

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationService.java
```diff
@@ -1,120 +1,71 @@
...
-import java.util.TreeMap;
-import java.util.logging.Logger;
 
 public class VerificationService implements com.sinch.sdk.domains.verification.VerificationService {
 
-  private static final Logger LOGGER = Logger.getLogger(VerificationService.class.getName());
-  // FIXME: Verification OAS file claim it support "Basic" but miss the "Application" definition
-  // trick to adapt the mapping of "Basic" keyword to the dedicated "Application" auth manager
-  private static final String BASIC_SECURITY_SCHEME_KEYWORD_VERIFICATION = "Basic";
+  private final com.sinch.sdk.domains.verification.api.v1.VerificationService v1;
```
asein-sinch, line 12: I'm a bit lost between all these `VerificationService`.
 - We have the interface (public, with the javadoc, exposed by the SinchClient) which is implemented by this adapter class.
 - And then the adapter delegates to the v1, which contains also an interface with the split start / report, implemented by another adapter class.

If I understand well, this class will stay during the transition period to do the redirection to the v1, but the endgame is to expose the v1 interface on the SinchClient, right? Or will this class always stay as a proxy?

pom.xml
JPPortier: Done

openapi-contracts/src/test/java/com/sinch/sdk/domains/verification/models/dto/v1/webhooks/VerificationResponseEventDtoTest.java
JPPortier: I asked myself the question but seems it could be an issue in case of a new field will be added Speech schema.
Imagine in this context a new field named "code" is added to the speech schema: it will cause conflicts with the "code" defined at root VerificationRequestEventResponsePhoneCall level.

Currently delegation is used in allOf context only  to create base class + inherited ones to "hide" and reduce subfield complexity. 
In this context we know we won't have conflicts by using delegation pattern

client/src/main/com/sinch/sdk/domains/verification/adapters/VerificationService.java
JPPortier: > If I understand well, this class will stay during the transition period to do the redirection to the v1, but the endgame is to expose the v1 interface on the SinchClient, right?

Totally right: this service is having to live during a transition period only, this to enable:
- smooth transition to "v1" pattern usage
- others products migration (SMS, Voice,... ) to support V1 migration too
So, finally: all products will use a common design: "the 'vN' usage enabling to support N different versions

asein-sinch (approved):
Impressive work with the new pattern while not breaking the current SDK interface 🌟

JPPortier:
> Impressive work with the new pattern while not breaking the current SDK interface 🌟

Thanks to SOC and data abstraction: it was the initial purpose of these adapters usage

PR #103, opened 2024-07-10
refactor: available/active classes moved to request/response package at numbers's package level

JPPortier:


asein-sinch (approved):
If I understand well how you have proceeded, only the classes in the packages 
 - client/src/main/com/sinch/sdk/domains/numbers/models/v1/request/*
 - client/src/main/com/sinch/sdk/domains/numbers/models/v1/response/*

 had to be moved manually?
 If that's the case, that's a huge benefit from the generator!

JPPortier:
> If I understand well how you have proceeded, only the classes in the packages
> 
>     * client/src/main/com/sinch/sdk/domains/numbers/models/v1/request/*
> 
>     * client/src/main/com/sinch/sdk/domains/numbers/models/v1/response/*
> 
> 
> had to be moved manually? If that's the case, that's a huge benefit from the generator!

Yes, in this context the "client" directory is not managed by generator; so moved manually yes.
The "automatically" moved files are the ones in "openapi-contracts/src/main" directory (and so generator handled automatically the proper "imports" dependencies within openapi-contracts/src/main files hierarchy)

PR #104, opened 2024-07-10
DEVEXP-238: synch with latest generator

JPPortier:
No added value here: just synchronization with generator output to prepare webhooks part


client/src/main/com/sinch/sdk/SinchClient.java
```diff
@@ -1,7 +1,7 @@
 package com.sinch.sdk;
 
 import com.sinch.sdk.core.utils.StringUtil;
-import com.sinch.sdk.domains.conversation.ConversationService;
+import com.sinch.sdk.domains.conversation.api.v1.ConversationService;
```
asein-sinch, line 4: This is the first full-fledge implementation of the new design! 🙌

asein-sinch (approved):

PR #114, opened 2024-08-09
fixes (Numbers): Fixes for OptionalValue getter and DelegatedSerializer

JPPortier:
- fix: OptionalValue getter for delegated fields wrongly returned OptionalValue.of(null) value instead of OptionalValue.empty() when delegated field was empty
- fix: Guard against null for DelegatedSerializer

### DIFF
--- a/openapi-contracts/src/main/com/sinch/sdk/domains/numbers/models/v1/errors/InternalErrorImpl.java
+++ b/openapi-contracts/src/main/com/sinch/sdk/domains/numbers/models/v1/errors/InternalErrorImpl.java
@@ -5,12 +5,20 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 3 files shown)


asein-sinch (approved):

openapi-contracts/src/main/com/sinch/sdk/domains/numbers/models/v1/errors/InternalErrorImpl.java
```diff
@@ -62,7 +70,9 @@ public String getMessage() {
   }
 
   public OptionalValue<String> message() {
-    return null != error ? error.map(InternalErrorError::getMessage) : OptionalValue.empty();
+    return null != error && error.isPresent()
+        ? error.map(f -> ((InternalErrorErrorImpl) f).message()).orElse(OptionalValue.empty())
```
asein-sinch, line 74: What is the reason why you now use the implementation instead of the interface to reach the property value?

JPPortier: The Impl is giving access to the `.message` getter which return an `OptionaValue`; then we know if value was set or not (and setting a value to null mean the value was set)
The `.getMessage` only return null (either the value was set to null or not initialized)
So the .getMessage is public to know if there is a value, but `.message` is internal to identify is value was set really or not.

PR #117, opened 2024-08-09
DEVEXP-307: Sync ContactMessage with updated specs (proposal)

JPPortier:


sample-app/src/main/java/com/sinch/sample/conversation/messages/Send.java
```diff
@@ -84,7 +84,7 @@ SendMessageRequest<?> createRCSSendMessage() {
         .setAppId(conversationAppId)
         .setMessage(
             AppMessage.<ChoiceMessage>builder()
-                .setMessage(
+                .setBody(
```
asein-sinch, line 87: 🤩

PR #118, opened 2024-08-12
Handle dynamic enums for url parameter

JPPortier:
Will be required for "Conversations list" where [filtering onto active channel](https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/Conversation/#tag/Conversation/operation/Conversation_ListConversations!in=query&path=active_channel&t=request)  require filtering onto an enum type to be transformed onto a String for URL parameters

### DIFF
--- a/core/src/test/java/com/sinch/sdk/core/http/URLParameterUtilsTest.java
+++ b/core/src/test/java/com/sinch/sdk/core/http/URLParameterUtilsTest.java
@@ -135,6 +135,22 @@ class URLParameterUtilsTest {
   }
 
   @Test
+  void encodeDynamicEnumStringNoExplode() {
+    assertEquals(
+        Optional.of("foo=ENUM1+value"),
+        URLParameterUtils.encode(
+            new URLParameter("foo", AnEnum.ENUM1, URLParameter.STYLE.FORM, false)));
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/sinch/sdk/core/http/URLParameterUtilsTest.java
```diff
@@ -134,6 +134,22 @@ void encodeFormStringExplode() {
...
+  @Test
+  void encodeDynamicEnumStringNoExplode() {
+    assertEquals(
+        Optional.of("foo=ENUM1+value"),
+        URLParameterUtils.encode(
+            new URLParameter("foo", AnEnum.ENUM1, URLParameter.STYLE.FORM, false)));
+  }
+
+  @Test
+  void encodeDynamicEnumStringExplode() {
```
asein-sinch: For these tests, having a couple of parameters would highlight the difference between the 2 values for the `explode` parameter.

JPPortier: Exact 👍 
Tests added

asein-sinch (approved):

PR #121, opened 2024-08-19
DEVEXP-536: Fix UTF-8 encoding for body payloads

JPPortier:


core/src/main/com/sinch/sdk/core/http/HttpContentType.java
```diff
@@ -17,4 +23,20 @@ public static boolean isMimeJson(Collection<String> mimes) {
...
+    if (StringUtil.isEmpty(contentTypeHeader)) {
+      return Optional.empty();
+    }
+
+    Matcher m = charsetPattern.matcher(contentTypeHeader);
+    if (!m.find()) {
+      return Optional.empty();
+    }
+    String group = m.group(2);
+    if (group.contains(";")) {
```
asein-sinch: You may update the regex to exclude the `;` and any space that may be present, such as `\\s*([^;\\s]+)`

JPPortier: Nice !
Done

asein-sinch (approved):

PR #126, opened 2024-08-20
feature (Conversation/Capability): 'Capability' for Conversation

JPPortier:


openapi-contracts/src/test/java/com/sinch/sdk/domains/conversation/models/v1/capability/request/QueryCapabilityRequestDtoTest.java
asein-sinch, line 1: Do you also have a unit test somewhere where you use a contactId as a recipient?

asein-sinch (approved):

openapi-contracts/src/test/java/com/sinch/sdk/domains/conversation/models/v1/capability/request/QueryCapabilityRequestDtoTest.java
JPPortier: At this dedicated test level: no.
Because of the request is accepting a _generic_ 'Recipient' instance; unit tests for serialize/deserialize a Recipient are here: https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-239-conversation-capability/openapi-contracts/src/test/java/com/sinch/sdk/domains/conversation/models/v1/RecipientDtoTest.java
So this test do not validate each use case by recipient inherited class but "only" capacity to serialize the request payload
Then it relies on RecipientDtoTest.java to provide full coverage about all Recipient's inherited classes

PR #132, opened 2024-08-26
DEVEXP-240: Conversation Templates V1 & V2

JPPortier:


sample-app/src/main/java/com/sinch/sample/conversation/templates/v2/Update.java
```diff
@@ -0,0 +1,72 @@
...
+                                TemplateVariable.builder()
+                                    .setKey("they key2")
+                                    .setPreviewValue("foo2")
+                                    .build()))
+                        .setChannelTemplateOverrides(
+                            Collections.singletonMap(
+                                ConversationChannel.WHATSAPP,
+                                ChannelTemplateOverride.builder()
+                                    .setParameterMappings(
+                                        Collections.singletonMap("they key1", "they key2"))
```
asein-sinch: I'm not sure this really makes sense to add parameters mapping when there is no template reference and thus no parameters for this template but I guess this is not the purpose of this test

openapi-contracts/src/test/resources/domains/conversation/templates/v1/TemplateTranslationDto.json
```diff
@@ -0,0 +1,13 @@
+{
+  "language_code": "fr-FR",
+  "content": "template translation content value",
```
asein-sinch: For educational purpose, it would be better to set a valid content for this parameter such as `{\"text_message\":{\"text\":\"Message from a template v1.\"}}`

sample-app/src/main/java/com/sinch/sample/conversation/templates/v2/Update.java
JPPortier: Right not important for this sample; but no more useful about the feature itself.
So by not adding this unused information it could avoid to misunderstood the feature.
Lines removed.

openapi-contracts/src/test/resources/domains/conversation/templates/v1/TemplateTranslationDto.json
JPPortier: Done

asein-sinch (approved):

PR #134, opened 2024-08-27
feat (SDK/Versioning): Store SDK related informations wihtin sources

JPPortier:
No longer using `.properties` file but store SDK related information at source level (client: [com.sinch.sdk.SDK.java](https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-510-store-sdk-information-with-sources/client/src/main/com/sinch/sdk/SDK.java))
Release procedure updated to synchronize `SDK.java`file content  onto github release action

Note: the release procedure require an Java 11 runtime enabling to _execute a java source file_: The JVM compiles the source file into memory and then runs the first public main() method it finds.
Then we are using [SDKTemplate Java source](https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-510-store-sdk-information-with-sources/client/resources/SDKTemplate.java) to update the `SDK.java` content, called from [release.sh script](https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-510-store-sdk-information-with-sources/scripts/release.sh#L29)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,8 +5,8 @@ on:
       releaseVersion:
         description: "Version number to be released (use semver semantic: 'X.Y.Z')"
         required: true
-      developmentVersion:
-        description: "Next development version to be set (use semver semantic: 'X.Y.Z', `-SNAPSHOT` suffix will be automatically added)"
+      nextDevelopmentVersion:
+        description: "Next development version to be set (use semver semantic: 'X.Y.Z', `-SNAPSHOT` artifact's suffix will be automatically added)"
         required: true
 
 permissions:
@@ -58,7 +58,7 @@ jobs:
         run: scripts/release.sh
         env:
           RELEASE_VERSION: ${{ github.event.inputs.releaseVersion }}
-          NEXT_VERSION: ${{ github.event.inputs.developmentVersion }}
+          NEXT_VERSION: ${{ github.event.inputs.nextDevelopmentVersion }}
           OSSRH_USER: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_PWD: ${{ secrets.OSSRH_PWD }}
           GPG_KEY_ID: ${{ secrets.GPG_KEY_ID }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -52,13 +52,13 @@ jobs:
 
       - name: Create and switch to dedicated release branch
         run: |
-          git checkout -b release-${{ github.event.inputs.releaseVersion }}
+          git checkout -b "release-${{ github.event.inputs.releaseVersion }}"
 
       - name: Release
-        run: |
-          mvn --batch-mode -s settings.xml release:clean release:prepare -Dtag=v${{ github.event.inputs.releaseVersion }} -DreleaseVersion=${{ github.event.inputs.releaseVersion }} -DdevelopmentVersion=${{ github.event.inputs.developmentVersion }}-SNAPSHOT -DscmReleaseCommitComment="[release] Set release & tag: ${{ github.event.inputs.releaseVersion }}" -DscmDevelopmentCommitComment="[release] Set next version: ${{ github.event.inputs.developmentVersion }}-SNAPSHOT"
-          mvn --batch-mode -s settings.xml release:perform -Dusername=${{ env.GITHUB_USERNAME }} -Dpassword=${{ env.GITHUB_TOKEN }} -DskipTests=true
+        run: scripts/release.sh
         env:
+          RELEASE_VERSION: ${{ github.event.inputs.releaseVersion }}
+          NEXT_VERSION: ${{ github.event.inputs.developmentVersion }}
           OSSRH_USER: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_PWD: ${{ secrets.OSSRH_PWD }}
           GPG_KEY_ID: ${{ secrets.GPG_KEY_ID }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -52,13 +52,13 @@ jobs:
...
       - name: Create and switch to dedicated release branch
         run: |
-          git checkout -b release-${{ github.event.inputs.releaseVersion }}
+          git checkout -b "release-${{ github.event.inputs.releaseVersion }}"
 
       - name: Release
-        run: |
-          mvn --batch-mode -s settings.xml release:clean release:prepare -Dtag=v${{ github.event.inputs.releaseVersion }} -DreleaseVersion=${{ github.event.inputs.releaseVersion }} -DdevelopmentVersion=${{ github.event.inputs.developmentVersion }}-SNAPSHOT -DscmReleaseCommitComment="[release] Set release & tag: ${{ github.event.inputs.releaseVersion }}" -DscmDevelopmentCommitComment="[release] Set next version: ${{ github.event.inputs.developmentVersion }}-SNAPSHOT"
-          mvn --batch-mode -s settings.xml release:perform -Dusername=${{ env.GITHUB_USERNAME }} -Dpassword=${{ env.GITHUB_TOKEN }} -DskipTests=true
+        run: scripts/release.sh
```
Dovchik, line 58: Is there is not available github action for the release?

client/src/main/com/sinch/sdk/SDK.java
```diff
@@ -0,0 +1,8 @@
+package com.sinch.sdk;
+
+public class SDK {
+
+  public static final String NAME = "Sinch Java SDK";
+  public static final String VERSION = "1.3.0-dev";
```
Dovchik, line 6: is this needs to be hardcoded?

scripts/release.sh
```diff
@@ -0,0 +1,47 @@
...
+  echo "Missing 'GITHUB_USERNAME' env value"
+  exit 1
+fi
+
+if [ ! "${GITHUB_TOKEN}"  ]; then
+  echo "Missing 'GITHUB_TOKEN' env value"
+  exit 1
+fi
+
+SDKFILE="client/src/main/com/sinch/sdk/SDK.java"
```
Dovchik: nit
```suggestion
SDKFILE_PATH="client/src/main/com/sinch/sdk/SDK.java"
```

client/resources/SDKTemplate.java
```diff
@@ -0,0 +1,20 @@
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class VersionTemplating {
+
+  public static void main(String[] args) throws IOException {
+    String versionFile = "client/src/main/com/sinch/sdk/SDK.java";
```
Dovchik: nit
```suggestion
    String versionFilePath = "client/src/main/com/sinch/sdk/SDK.java";
```

client/resources/SDKTemplate.java
```diff
@@ -0,0 +1,20 @@
...
+
+    Stream<String> lines = Files.lines(Paths.get(versionFile));
+    String content = lines.collect(Collectors.joining("\n"));
+    lines.close();
+
+    System.out.println(
+        content.replaceAll("VERSION = \".*\"", String.format("VERSION = \"%s\"", args[0])));
```
Dovchik: Would be nice to see `args[0]` extracted into variable with meaningful name, will spare 5 secs of thinking
```suggestion
        content.replaceAll("VERSION = \".*\"", String.format("VERSION = \"%s\"", args[0])));
```

.github/workflows/release.yaml
```diff
@@ -52,13 +52,13 @@ jobs:
...
         env:
+          RELEASE_VERSION: ${{ github.event.inputs.releaseVersion }}
+          NEXT_VERSION: ${{ github.event.inputs.developmentVersion }}
```
Dovchik: Why is that `next_version` and `dev version` would it be more consistent to have one name for that?

.github/workflows/release.yaml
JPPortier: It could but when it is time to sending credentials to an external plugin I'm in zero trust mood

PROS for home made:
- home made recipes not sending credentials (`mvn release` is itself a mvn plugin, but why to exposing credentials twice if only one is enough and [sources are public](https://github.com/apache/maven-release/) )
- full control onto our script if fixes or enhancements are required

CONS for external action:
- tie to a 3rd party plugin 
- what if maintainer do not respond/provide support
- what if plugin is no longer maintained

client/src/main/com/sinch/sdk/SDK.java
JPPortier: This is the purpose of the PR (see description for details about how it is updated)

scripts/release.sh
JPPortier: Right.
Done

client/resources/SDKTemplate.java
JPPortier: Right.
Changed with `SDK_FILE_PATH`

client/resources/SDKTemplate.java
JPPortier: Thank you.
This is effectively a better implementation
Done.

.github/workflows/release.yaml
JPPortier: Yes, renamed github action's input variable from `developmentVersion` to `nextDevelopmentVersion`
More consistent with input description: `"Next development version to be set (use semver semantic: 'X.Y.Z', -SNAPSHOT artifact's suffix will be automatically added)"`

.github/workflows/release.yaml
Dovchik: In my test I like not to reinvent the wheel but it's up to.

client/src/main/com/sinch/sdk/SDK.java
Dovchik: nit
maybe put comment or stub value to indicate that it's overriden from external executor

Dovchik (approved):

asein-sinch (approved):

PR #135, opened 2024-08-28
DEVEXP-530: Conversation e2e tests (Apps)

JPPortier:
Based onto Cucumber and Sinch SDK MockServer: e2e for Conversation/Apps
(`build.yml` appear as a new file but it was renamed and enhanced)

Notes: runing tests in multithreading mode based onto a single SinchClient instance

Tests output:
```
[INFO] Running com.sinch.sdk.e2e.domains.conversation.ConversationIT
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.449 s -- in com.sinch.sdk.e2e.domains.conversation.ConversationIT
```


client/src/test/java/com/sinch/sdk/e2e/domains/conversation/AppsSteps.java
```diff
@@ -0,0 +1,140 @@
...
+        ChannelIntegrationStatus.ACTIVE);
+  }
+
+  @Then("the delete app response contains no data")
+  public void deleteResult() {
+
+    Assertions.assertTrue(deletePassed);
+  }
+
+  void checkExpectedAppResponseCommonFields(AppResponse appResponse) {
```
asein-sinch, line 133: This is a nice idea to factorize the check on common fields. But some tests can break if we change some data in the mockserver expectations. I've tried to create some consistency when I wrote them. Let's agree together to keep it like this!

client/src/test/java/com/sinch/sdk/e2e/domains/conversation/AppsSteps.java
```diff
@@ -0,0 +1,140 @@
...
+    AppUpdateRequest request =
+        AppUpdateRequest.builder().setDisplayName(" a new display name").build();
+    updateResponse = service.update("foo", request);
+  }
+
+  @When("^I send a request to delete an app$")
+  public void delete() {
+
+    service.delete("foo");
+    deletePassed = true;
```
asein-sinch, line 83: Here you are just checking that the request completed without exception right?
Note that if we decide to move to static values instead of regexp for parameter matching, this test will fail as `foo` is certainly not a value corresponding to an appId. You may reuse the one tou are already checking in the common fields

asein-sinch (approved):

client/src/test/java/com/sinch/sdk/e2e/domains/conversation/AppsSteps.java
JPPortier: According to our discussion: we agree to update tests accordingly in case of future change

client/src/test/java/com/sinch/sdk/e2e/domains/conversation/AppsSteps.java
JPPortier: According to our discussion: we agree to update tests accordingly in case of future change

PR #139, opened 2024-08-30
DEVEXP-530: Conversation e2e (alls but templates)

JPPortier:
- Note yet templates related test due to pending Temapltes PR in review (will come after)

- @asein-sinch About https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-530-conversation-e2e/client/src/test/java/com/sinch/sdk/e2e/domains/conversation/WebhooksEventsSteps.java#L282
We can imagine refining mockerserver/cucumber description with something like `the Conversation event header contains a valid signature for {string}` 
Then tests could use the `{string}` identifier to detect which event is having to be checked when running multihtread/parallel mode


JPPortier:
@asein-sinch Created DEVEXP-550 regarding the authentication header validation

asein-sinch (approved):

PR #146, opened 2024-09-25
DEVEXP-574: Conversation credentials refining

JPPortier:
Use shared `ConversationChannelCredentials` for both get/set requests


client/src/test/java/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsBuilderFactoryTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+import com.sinch.sdk.core.TestHelpers;
+import com.sinch.sdk.domains.conversation.models.v1.ConversationChannel;
+import org.junit.jupiter.api.Test;
+
+class ConversationChannelCredentialsBuilderFactoryTest {
+
+  @Test
+  void appleBc() {
+    AppleBusinessChatCredentials credentials = AppleBusinessChatCredentials.builder().build();
```
asein-sinch, line 13: Nit: even if this code is 100% legit, from a business point of view, it doesn't make sense to create a credentials object without the property that is declared mandatory in the sub-object related to the channel's credentials.

asein-sinch (approved):

client/src/test/java/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsBuilderFactoryTest.java
JPPortier: Yes: the purpose of these unit tests is related to BuilderFactory business logic validation; it do not check credentials content (this is done with these unit tests: https://github.com/sinch/sinch-sdk-java/blob/a5a64435a8978b76b0c2632e9af26597f2f3be18/openapi-contracts/src/test/java/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsDtoTest.java)

PR #147, opened 2024-09-26
DEVEXP-542: Conversation javadoc

JPPortier:


client/src/main/com/sinch/sdk/domains/conversation/api/v1/MessagesService.java
```diff
@@ -49,22 +49,85 @@ public interface MessagesService {
    */
   SendMessageResponse sendMessage(SendMessageRequest<? extends AppMessageBody> request);
 
+  /**
+   * Send a card message
+   *
+   * @param request CardMessage request
+   * @return Response
```
asein-sinch: This javadoc is strange:
 * for the parameter, the word `request` is repeated and `CardMessage` is not really the class to use
 * for the return, the response is a `SendMessageResponse` instead of a `Response`

client/src/main/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsBuilderFactory.java
```diff
@@ -2,90 +2,191 @@
...
 public class ConversationChannelCredentialsBuilderFactory {
 
+  private ConversationChannelCredentialsBuilderFactory() {}
+
+  /**
+   * Create dedicated conversation credential builder for <code>{@link ConversationChannel#APPLEBC}
+   * </code> channel
+   *
+   * @param credentials Credentials
+   * @return Credentials builder
```
asein-sinch: Review classes here too:
 - @param:  Credentials -> AppleBusinessChatCredentials
 - @return: Credentials -> ConversationChannelCredentials

asein-sinch (approved):

client/src/main/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsBuilderFactory.java
JPPortier: Same as previous comment: mix between parameter name and used type.
Changed to have better doc
image

client/src/main/com/sinch/sdk/domains/conversation/api/v1/MessagesService.java
JPPortier: > This javadoc is strange:
>     * for the parameter, the word `request` is repeated and `CardMessage` is not really the class to use

Yes, here the 1st 'request' is the parameter name from function prototype. Then `CardMessage request` is the message used for documentation
javadoc output:
image

> for the return, the response is a SendMessageResponse instead of a Response

Purpose here was not document return type (javadoc do it automatically from prototype)
Changed in favor of a true sentence
image

client/src/main/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsBuilderFactory.java
asein-sinch: Thanks. Too much jsdoc recently on my side 😉

PR #162, opened 2024-10-02
refactor (samples): Use webhooks at request level when defined

JPPortier:


sample-app/src/main/java/com/sinch/sample/conversation/messages/Send.java
```diff
@@ -119,5 +126,8 @@ SendMessageRequest<?> createSMSSendMessage() {
                     .build()))
         .setTtl(25)
         .build();
+
+    webhooksConversationPath.ifPresent(builder::setCallbackUrl);
```
asein-sinch, line 130: Suggestion: wouldn't is be clearer to use the setter in the builder chain above?
```
.setCallbackUrl(webhooksConversationPath.orElse(null));
```

sample-app/src/main/java/com/sinch/sample/verification/start/Start.java
```diff
@@ -34,7 +34,7 @@ public void run() {
     LOGGER.info("Start verification for : " + phoneNumber);
 
     NumberIdentity identity = NumberIdentity.valueOf(phoneNumber);
-    VerificationMethod method = VerificationMethod.FLASH_CALL;
+    VerificationMethod method = VerificationMethod.SMS;
```
asein-sinch, line 37: 💬 vs 📞

sample-app/src/main/java/com/sinch/sample/conversation/messages/Send.java
JPPortier: no: this will lead to send the `null` value as the `callback_url` value
Setting a value (even null) mean: "send the value"

asein-sinch: Arf, that's true. So you are forced to do this manipulation to use an optional parameter in the builder whose value may come or be empty from the configuration file 😣

asein-sinch (approved):

sample-app/src/main/java/com/sinch/sample/conversation/messages/Send.java
JPPortier: Yes. By purpose.
We are in samples here: if configuration file define a value it supersed the default from dashboard (if any)
If not set, not sent and use dashboard configuration
Again: just a matter of samples; SDK itself is not in the loop

PR #169, opened 2024-11-26
DEVEXP-648: Form parameters serializer

JPPortier:
1. Support form parameters serialization by introspection enabling dedicated serializer at field level definition
  - `Instant` to `RFC822` string for delivery time
  - XX `Integer` value to "XX`H`" string for  delivery time optimize period
2. additionalProperties is not yet supported and be part of dedicated PR


core/src/main/com/sinch/sdk/core/databind/multipart/ObjectMapper.java
```diff
@@ -0,0 +1,127 @@
...
+      return properties;
+    }
+
+    ArrayList<Pair<String, Method>> sorted = new ArrayList<>(properties.size());
+
+    for (String property : propertyOrder.value()) {
+      properties.stream()
+          .filter(p -> p.getLeft().equals(property))
+          .findFirst()
+          .ifPresent(sorted::add);
```
asein-sinch: We can be more efficient here: you are repeatedly filtering through the properties list (O(n) for each property in propertyOrder). Instead, you can create a Map<String, Method> for faster lookups (O(1))

```suggestion
    // Convert the properties list to a map for efficient lookups
    Map<String, Method> propertyMap = properties.stream()
            .collect(Collectors.toMap(Pair::getLeft, Pair::getRight));

    List<Pair<String, Method>> sorted = Arrays.stream(propertyOrder.value())
            .map(property -> Pair.of(property, propertyMap.get(property)))
            .filter(pair -> pair.getRight() != null)
            .collect(Collectors.toList());
```

And I'm not sure to understand how you handle the additional properties, but you then can add the following code to add them at the end of the sorted list:
```java
properties.stream()
    .filter(p -> !sorted.stream().anyMatch(s -> s.getLeft().equals(p.getLeft())))
    .forEach(sorted::add);
```

core/src/test/java/com/sinch/sdk/core/databind/multipart/ObjectMapperTest.java
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  public static class RFC822ListFormSerializer extends FormSerializer<Collection<Instant>> {
+
+    @Override
+    public Collection<String> serialize(Collection<Instant> in) {
+      return in.stream()
+          .map(
+              instant ->
+                  DateTimeFormatter.RFC_1123_DATE_TIME.format(instant.atZone(ZoneId.of("UTC"))))
```
asein-sinch: I would have expected to see a reference to `RFC822FormSerializer` here

core/src/test/java/com/sinch/sdk/core/databind/multipart/ObjectMapperTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+  @PropertiesOrder({
+    SerializableObject.PROPERTY_RFC822_COLLECTION,
+    SerializableObject.PROPERTY_TEXT,
+    SerializableObject.PROPERTY_FILE_COLLECTION,
+    SerializableObject.PROPERTY_RFC822,
+    SerializableObject.PROPERTY_FILE,
+    SerializableObject.PROPERTY_ENUM,
+    SerializableObject.PROPERTY_ENUM_COLLECTION,
+    SerializableObject.PROPERTY_RFC822_COLLECTION,
```
asein-sinch: Is it expected to have this property duplicated?

core/src/test/java/com/sinch/sdk/core/databind/multipart/ObjectMapperTest.java
JPPortier: Yes it do the "same" job:
- https://mailgun-docs.redoc.ly/docs/mailgun/api-reference/#date-format
- From javadoc: : `RFC_1123_DATE_TIME` handle RFC 1123 / RFC 822 with 4 years digit (https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#RFC_1123_DATE_TIME)

And the four digit for years is what is pointed by link from Mailgun documentation: https://datatracker.ietf.org/doc/html/rfc2822.html#page-14
```
year            =       4*DIGIT / obs-year
```

core/src/test/java/com/sinch/sdk/core/databind/multipart/ObjectMapperTest.java
JPPortier: Absolutely not.
Fixed within this PR/commit: https://github.com/sinch/sinch-sdk-java/pull/171/commits/d45ce941c4e56f52515c035439f129c7236277ce

core/src/main/com/sinch/sdk/core/databind/multipart/ObjectMapper.java
JPPortier: Thank you to highlight the optimization !
Choose this implementation in place of you suggestion: https://github.com/sinch/sinch-sdk-java/pull/171/commits/b2d7acb077439ed9a113463befb0b5de4e880a75
It avoid to:
- create a new stream to create a map the collected to a list -> No need for 
```
    Map<String, Method> propertyMap = properties.stream()
            .collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
```
- create a stream then Pairs to be collected to a List -> No need for 
```
    List<Pair<String, Method>> sorted = Arrays.stream(propertyOrder.value())
            .map(property -> Pair.of(property, propertyMap.get(property)))
            .filter(pair -> pair.getRight() != null)
            .collect(Collectors.toList());
```

Finally; only have:
```
    List<String> order = java.util.Arrays.asList(propertiesOrder.value());
    properties.sort(Comparator.comparingInt(l -> order.indexOf(l.getLeft())));
```
Still having to create a List from Array to have `order` variable
But the sort operation is performed directly onto the list received as parameter

asein-sinch: No more sorting, no more processing ✅

core/src/test/java/com/sinch/sdk/core/databind/multipart/ObjectMapperTest.java
JPPortier: 👌 
Refactored here: https://github.com/sinch/sinch-sdk-java/pull/171/commits/d075b64e8d1a33e950e1d054b98bc3dbe50b3994

asein-sinch (approved):

PR #176, opened 2024-12-16
DEVEXP-661: use url parameter for purge domain

JPPortier:


asein-sinch (approved):
Approved, even in the PR name should be "purge sending queue" and not "purge domain" 😄

PR #180, opened 2024-12-18
test (SMS/Batches): Adding batches API calls regression tests

JPPortier:


client/src/test/java/com/sinch/sdk/domains/sms/api/v1/internal/BatchesApiTest.java
```diff
@@ -0,0 +1,332 @@
...
+            Collections.singletonList(SMS_AUTH_NAMES));
+    HttpResponse httpResponse = new HttpResponse(200, null, Collections.emptyMap(), null);
+
+    when(httpClient.invokeAPI(
+            eq(serverConfiguration),
+            eq(authManagers),
+            argThat(new HttpRequestMatcher(httpRequest))))
+        .thenReturn(httpResponse);
+
+    api.sendDeliveryFeedback("foo text batch id", sendDeliveryFeedbackRequestDto);
```
asein-sinch, line 330: Isn't there anything that can be checked to make sure the serialization was ok and the parameters are matching?

JPPortier: The _beauty_ is here: `argThat(new HttpRequestMatcher(httpRequest))))`
1. `HttpRequestMatcher` (https://github.com/sinch/sinch-sdk-java/pull/180/files#diff-585489e9fb9cb6a5f6d4c8764f52f2fb7407d1ef0b8cdee3107fa0a6fbe871b6R18) will do the validation against expected argument by validating class type and values (recursively)
2. if it match: it will return `true`: this mean for mockito that the argument is having the expected content. If it do not match -> Exception with details
3. Finally, even if not using `HttpRequestMatcher` mockito itself would/will  raise an exception (but without details) because he do not find a stub describing the expected arguments described from:
```
    when(httpClient.invokeAPI(
            eq(serverConfiguration),
            eq(authManagers),
            argThat(new HttpRequestMatcher(httpRequest))))
```

To summarize:an exception will be throw by `HttpRequestMatcher` if serialized HttpRequest do not match (the aim of these tests)

**Sample of outputs**
- when changing expected URL to:
```
        new HttpRequest(
            "/xms/v1/foovalue/batches/foo%20text%20batch%20id/WRONGENDPOINT",
```
Result:
```
java.lang.AssertionError: 
Expecting actual:
  com.sinch.sdk.core.http.HttpRequest@421def93
to be equal to:
  com.sinch.sdk.core.http.HttpRequest@58c1da09
when recursively comparing field by field, but found the following difference:

field/property 'path.value' differ:
- actual value  : "/xms/v1/foovalue/batches/foo%20text%20batch%20id/delivery_feedback"
- expected value: "/xms/v1/foovalue/batches/foo%20text%20batch%20id/WRONGENDPOINT"
...
```
- when changing an expected field from payalod (1st phone number changed to: `+15551231234XXX`:
```
java.lang.AssertionError: 
Expecting actual:
  com.sinch.sdk.core.http.HttpRequest@421def93
to be equal to:
  com.sinch.sdk.core.http.HttpRequest@58c1da09
when recursively comparing field by field, but found the following difference:

field/property 'body' differ:
- actual value  : "{"recipients":["+15551231234","+15987365412"]}"
- expected value: "{"recipients":["+15551231234XXX","+15987365412"]}"
...
```

asein-sinch: Thanks for the explanation!

asein-sinch (approved):

PR #182, opened 2025-01-09
DEVEXP-469: delivery reports v1

JPPortier:


client/src/main/com/sinch/sdk/domains/sms/api/v1/adapters/DeliveryReportsService.java
```diff
@@ -0,0 +1,64 @@
...
+      String uriUUID,
+      SmsContext context,
+      HttpClient httpClient,
+      Map<String, AuthManager> authManagers) {
+    this.api =
+        new DeliveryReportsApi(
+            httpClient, context.getSmsServer(), authManagers, new HttpMapper(), uriUUID);
+  }
+
+  public BatchDeliveryReport get(String batchId, BatchDeliveryReportQueryParameters parameters)
```
asein-sinch, line 43: In this adapter, would it also be possible to create another method `get(String batchId)` in order to no force the user to add a null parameter when no optional query parameters are needed?

JPPortier: Totally.
Investigating to handle this case from sources generator tool

asein-sinch: As you want, but my suggestion was to add it inside the adapter as an "elegant" function

JPPortier: Yes, but finally because want to move away from adapters, if managed at generator level it will be better
Additionally: we can imagine more functions with only one or two parameters across different endpoints. So not trigerring a complex class with very few fields and builder should improve DX

asein-sinch (approved):

PR #183, opened 2025-01-09
DEVEXP-621: mailgun messages support dedicated subfields

JPPortier:


client/src/test/java/com/sinch/sdk/domains/mailgun/api/v1/adapters/CustomHeadersFormSerializerTest.java
```diff
@@ -0,0 +1,49 @@
...
+  void serialize() {
+
+    HashMap<String, Object> map = new HashMap<>();
+
+    serializer.serialize(
+        Arrays.asList(
+            Pair.of("my-header-key1", "a-header-value1"),
+            Pair.of("my-header-key1", "a-header-value2"),
+            Pair.of("my-header-key2", "a-header-value3")),
+        "foo name",
```
asein-sinch: This is not to be used by the end user, but we can think about adding a new signature for the method `serialize`. This new method would be used in cases like this one, where the parameter key is extracted from the values

client/src/main/com/sinch/sdk/domains/mailgun/api/v1/adapters/NestedFieldFormSerializer.java
```diff
@@ -0,0 +1,19 @@
+package com.sinch.sdk.domains.mailgun.api.v1.adapters;
+
+import com.sinch.sdk.core.databind.FormSerializer;
+import com.sinch.sdk.core.utils.databind.MultiPartMapper;
+import java.util.Map;
+
+public class NestedFieldFormSerializer extends FormSerializer<Object> {
```
asein-sinch, line 7: Can you explain me what this serializer is used for?

core/src/main/com/sinch/sdk/core/utils/Pair.java
```diff
@@ -39,6 +39,19 @@ public V getRight() {
     return this.right;
   }
 
+  /**
+   * Helper to create new instance of Pair
+   *
+   * @param key key value
+   * @param value value of value
```
asein-sinch, line 46: Indeed 🤣

openapi-contracts/src/main/com/sinch/sdk/domains/mailgun/models/v1/emails/request/TemplateProperties.java
```diff
@@ -0,0 +1,86 @@
...
+  String getVersion();
+
+  /**
+   * A valid JSON-encoded dictionary used as the input for template variable expansion. See <a
+   * href="https://documentation.mailgun.com/docs/mailgun/user-manual/sending-messages/#templates">Templates</a>
+   * for more information
+   *
+   * @return variables
+   */
+  String getVariables();
```
asein-sinch, line 38: I missed this in the previous review: we can enhance this property too to be a Record<string, something>, the `something` being a `string` | `boolean` | `Date` | `number`

asein-sinch: (but to be done in the generator first, this is not blocking this PR)

client/src/test/java/com/sinch/sdk/domains/mailgun/api/v1/adapters/CustomHeadersFormSerializerTest.java
JPPortier: 🤔 ... yes but no....
- Yes: Even if used internally -> better self serve DX 😃 
- No: Currently this "feature" is only required from serializers unit tests themselves

So adding code to satisfy only tests purpose only could be seen as a bad practice
In real life "use case" they're will always a defined key to be mapped to  
So, won't change the interface but changing the test to use "null" in place of a "foo" key. It will highlight+ensure "null" is well supported even if never occur in real life (hello Murphy!)

client/src/main/com/sinch/sdk/domains/mailgun/api/v1/adapters/NestedFieldFormSerializer.java
JPPortier: Used for override & templates properties:
- https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-621-mailgun-messages-support-dedicated-subfields/openapi-contracts/src/main/com/sinch/sdk/domains/mailgun/models/v1/emails/request/SendEmailHtmlInTemplateRequestImpl.java#L209
- https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-621-mailgun-messages-support-dedicated-subfields/openapi-contracts/src/main/com/sinch/sdk/domains/mailgun/models/v1/emails/request/SendEmailHtmlInTemplateRequestImpl.java#L251

(Explanation for OverrideProperties but same logic for TemplateProperties.)

Override properties is a dedicated class with it own serialization to change all fields names `xxx` to `o:xxx` (e.g https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-621-mailgun-messages-support-dedicated-subfields/openapi-contracts/src/main/com/sinch/sdk/domains/mailgun/models/v1/emails/request/OverridePropertiesImpl.java#L22-L24)
So serialization of OverrideProperties serialization produce the expected prefixed values BUT OverrideProperties is a subfield of message : https://github.com/sinch/sinch-sdk-java/blob/DEVEXP-621-mailgun-messages-support-dedicated-subfields/openapi-contracts/src/main/com/sinch/sdk/domains/mailgun/models/v1/emails/request/SendEmailHtmlInlineRequestImpl.java#L57-L59

So if we do not set a dedicated serializer for `SendEmailHtmlInlineRequest.overrideProperties` field; serialization will produce a subfield content to be stored in map under the `overrideProperties` key.
What is expected is that all the `o:xxx` fields are stored in map at same level as all fields (no hierarchy+depth).
This is role of NestedFieldFormSerializer it take all fields from serialized OverrideProperties to store them one level up:  root of output
So finally we have 'flattened' the subfields 🏅

core/src/main/com/sinch/sdk/core/utils/Pair.java
JPPortier: 👼

openapi-contracts/src/main/com/sinch/sdk/domains/mailgun/models/v1/emails/request/TemplateProperties.java
JPPortier: Good catch.
Will be part of a dedicated feature's PR from generator 1st and side effects onto Node & Java)

asein-sinch (approved):

PR #184, opened 2025-01-10
DEVEXP-469: webhooks v1

JPPortier:


client/src/main/com/sinch/sdk/domains/sms/api/v1/WebHooksService.java
```diff
@@ -0,0 +1,9 @@
+package com.sinch.sdk.domains.sms.api.v1;
+
+import com.sinch.sdk.core.exceptions.ApiMappingException;
+import com.sinch.sdk.domains.sms.models.v1.webhooks.SmsEvent;
+
+public interface WebHooksService {
+
+  SmsEvent parse(String jsonPayload) throws ApiMappingException;
```
asein-sinch: Should be `parseEvent` according to what is done in other domains

client/src/main/com/sinch/sdk/domains/sms/models/v1/deliveryreports/DeliveryReport.java
asein-sinch, line 1: Is there a benefit to add an extra layer for the `BatchDeliveryReport` and `RecipientDeliveryReport` or could they extend `SmsEvent` directly?

client/src/main/com/sinch/sdk/domains/sms/api/v1/WebHooksService.java
JPPortier: Right. Fixed

client/src/main/com/sinch/sdk/domains/sms/models/v1/deliveryreports/DeliveryReport.java
JPPortier: It could, both of them will identical in term of supported features.
But 
- by having a common `DeliveryReport`, it become consistent with existing `InboundMessage` pattern now also extending `SmsEvent`
- in future if a new report is having to be supported (so  same for Inbound) as soon have it become a `DeliveryReport` it become also a _SmsEvent_ by inheritance

asein-sinch (approved):

PR #186, opened 2025-01-14
DEVEXP-702:  SMS webhooks validate authentication headers

JPPortier:
~~Will require mockserver's PR #62 merged before e2e passed~~

Extra steps are not executed because of not yet available from mockserver; so tests passed because not yet fully executed.


asein-sinch (approved):
LGTM 🌟

PR #187, opened 2025-01-14
DEVEXP-469: SMS groups v1

JPPortier:


client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/GroupsServiceTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+    TestHelpers.recursiveEquals(response, groupResponseDto);
+  }
+
+  @Test
+  void create() throws ApiException {
+
+    when(api.createGroup(eq(null))).thenReturn(groupResponseDto);
+
+    Group response = service.create(null);
```
asein-sinch: When passing `null` as a body parameter, does it generate an empty body that will be sent on the network? I'm asking this question because the server won't accept the request if there is no body

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/GroupsServiceTest.java
```diff
@@ -0,0 +1,216 @@
...
+            .setSize(2)
+            .setCreatedAt(Instant.parse("2019-08-24T14:15:22Z"))
+            .setModifiedAt(Instant.parse("2019-08-24T14:15:44Z"))
+            .setAutoUpdate(
+                GroupAutoUpdate.builder()
+                    .setTo("15551231234")
+                    .setAdd(AddKeyword.builder().setFirstWord("Add 1st keyword").build())
+                    .setRemove(
+                        RemoveKeyword.builder()
+                            .setFirstWord("remove 1st keyword")
```
asein-sinch: Seen in the documentation: `Constraints: Must be one word.`
Also seen: `string <= 15 characters`
I know this is not the purpose of this test to use valid inputs, but... still trying to make you change your mind 😝

openapi-contracts/src/test/java/com/sinch/sdk/domains/sms/models/v1/groups/request/GroupUpdateRequestDtoTest.java
```diff
@@ -0,0 +1,35 @@
...
+  String jsonGroupUpdateRequestDto;
+
+  @Test
+  void serialize() throws JsonProcessingException, JSONException {
+
+    GroupUpdateRequest requestDTO =
+        GroupUpdateRequest.builder()
+            .setName("My new customers")
+            .setAdd(Arrays.asList("foo"))
+            .setRemove(Arrays.asList("01FC66621XXXXX119Z8PMV1AHY", "01FC66621XXXXX119Z8PMV1A00"))
```
asein-sinch: Well, here too the data is not meaningful. It's even confusing: for the `remove` property, it should be a list of phone numbers and in the `remove_from_group` a list a group IDs.

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/GroupsServiceTest.java
JPPortier: Exact.
And OAS specs do not claim it as `required` (by default body are optional)
So finally, this test do not have sense.
Removing it.
(and: no there is no helper to provide an empty body if null... should be part of a cross domains ticket to be consistent for all of them)

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/GroupsServiceTest.java
JPPortier: > till trying to make you change your mind 😝
🤔 ..... OK, you won
Changed (not 'fixed' just 'changed' 😜 )

openapi-contracts/src/test/java/com/sinch/sdk/domains/sms/models/v1/groups/request/GroupUpdateRequestDtoTest.java
JPPortier: Changed

asein-sinch (approved):

PR #190, opened 2025-01-17
Devexp 706 sms services adaptor removed

JPPortier:


client/src/test/java/com/sinch/sdk/domains/PaginationFillerHelper.java
```diff
@@ -0,0 +1,29 @@
...
+import com.sinch.sdk.core.http.URLParameter;
+import com.sinch.sdk.core.http.URLParameter.STYLE;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class PaginationFillerHelper {
+
+  public static Collection<URLParameter> parametersFiller(
+      String name, Object value, STYLE style, boolean explode, List<?> values) {
```
asein-sinch, line 12:
```java
List<?> values
```
Why not? 😁
This is not the most elegant way, but it works

openapi-contracts/src/main/com/sinch/sdk/domains/sms/models/v1/batches/response/internal/ApiBatchList.java
```diff
@@ -35,9 +35,9 @@ public interface ApiBatchList {
   /**
    * The page of batches matching the given filters.
    *
-   * @return batches
+   * @return items
    */
-  List<BatchResponse> getBatches();
+  List<BatchResponse> getItems();
```
asein-sinch, line 40: Regarding the fact that the getters and setters have changed for all the `list` requests, does it have an impact for the SDK user? (I've noticed that these changes are limited to the `internal` package, but I wanted to double-check with you)

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/BatchesServiceTest.java
```diff
@@ -1,515 +1,481 @@
...
+    ListBatchesResponse response = service.list(initialRequest);
 
-    TestHelpers.recursiveEquals(response, batchBinary);
-  }
+    Iterator<BatchResponse> iterator = response.iterator();
 
-  @Test
-  void replaceBinary() throws ApiException {
+    BatchResponse item = iterator.next();
+    TestHelpers.recursiveEquals(item, listBatchesResponseDtoPage0.getItems().get(0));
```
asein-sinch, line 367: This is on this `.getItems()` method that I wanted to double-check there is not an impact on the SDK user

client/src/test/java/com/sinch/sdk/domains/PaginationFillerHelper.java
JPPortier: Right. 
No need of generics here, an simple Object was enough, and for tests purpose only I did not move to implement an typed object... the idea if to avoid URLParameters typed object usage to have simplified calls.
This is a tests helper only; and if List is filled wrong... then the test itself will fails

openapi-contracts/src/main/com/sinch/sdk/domains/sms/models/v1/batches/response/internal/ApiBatchList.java
JPPortier: No breaking change here: the ApiBacthList is for internal purpose only and the previous "not versioned" design to do not use it but provide adaptors 
So:
- either the user is still using adaptors design -> no BC
- either he move to versioned design: and it is not a breaking change because all the design is new and there is backward compability with adaptors design

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/BatchesServiceTest.java
JPPortier: 🆗 see above: only for new versioned design usage and internal use only

asein-sinch (approved):

PR #192, opened 2025-01-20
Mailgun's javadoc

JPPortier:


client/src/main/com/sinch/sdk/domains/mailgun/MailgunService.java
```diff
@@ -4,7 +4,7 @@
  * Mailgun Service
  *
  * @see <a
- *     href="https://documentation.mailgun.com/docs/mailgun/">https://documentation.mailgun.com/docs/mailgun</a>
+ *     href="https://documentation.mailgun.com/docs/mailgun/user-manual/get-started">https://documentation.mailgun.com/docs/mailgun/user-manual/get-started/</a>
```
asein-sinch, line 7: Note: this SDK is not the official one yet and the linked documentation refers to the other one. Let's keep it this way for the moment, but when it will be released, we will have to request an update to the documentation team.

asein-sinch (approved):

client/src/main/com/sinch/sdk/domains/mailgun/MailgunService.java
JPPortier: right.
There is pending task from Doc team about this subject
A lot of links will have to be check/reviewed in the future

PR #193, opened 2025-01-20
fix: adjust channel_specific_message model

Dovchik:
Maybe needed model adjust


JPPortier:
Closing this request.
Will require more work with related opened ticket

PR #197, opened 2025-01-23
DEVEXP-719: cross domains deprecation notice

JPPortier:


client/src/main/com/sinch/sdk/domains/sms/GroupsService.java
```diff
@@ -47,15 +54,21 @@ public interface GroupsService {
...
    */
+  @Deprecated
   Group create(GroupCreateRequestParameters parameters) throws ApiException;
 
   /**
    * Create an unnamed and empty group
    *
    * @return See {@link #create(GroupCreateRequestParameters)}
    * @since 1.0
+   * @deprecated use {@link com.sinch.sdk.domains.sms.api.v1.GroupsService#create(GroupRequest)}
```
asein-sinch, line 68: Is it intended to remove the support of the empty group creation helper?

JPPortier: No.
You're right this helper was lost in translation...
I created a ticket: mainly to be able to generate such helper out of the box from generator when all parameters for a request will be optional: DEVEXP-728

asein-sinch (approved):

PR #206, opened 2025-02-10
DEVEXP-736: relax credentials required for webhooks conversation

JPPortier:
Enable init and use a SinchClient instance without credential
It enable header authentication validation and event parsing not requiring credentials

### DIFF
--- a/client/src/main/com/sinch/sdk/domains/conversation/api/templates/adapters/TemplatesService.java
+++ b/client/src/main/com/sinch/sdk/domains/conversation/api/templates/adapters/TemplatesService.java
@@ -15,6 +15,7 @@ import com.sinch.sdk.models.UnifiedCredentials;
 import java.util.AbstractMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Supplier;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -26,11 +27,15 @@ public class TemplatesService
   private static final String SECURITY_SCHEME_KEYWORD_ = "oAuth2";
 
   private final String uriUUID;
+  private final UnifiedCredentials credentials;
+  private final ServerConfiguration oAuthServer;
+
   private final ConversationContext context;
-  private final HttpClient httpClient;
-  private final Map<String, AuthManager> authManagers;
-  private TemplatesServiceV1 v1;
-  private TemplatesServiceV2 v2;
+  private final Supplier<HttpClient> httpClientSupplier;
+
+  private volatile Map<String, AuthManager> authManagers;
+  private volatile TemplatesServiceV1 v1;
+  private volatile TemplatesServiceV2 v2;
 
   static {
     LocalLazyInit.init();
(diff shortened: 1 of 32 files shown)


asein-sinch (commented):
Just wondering: what's the benefit to tie the webhook validation with a SinchClient? Wouldn't a static method be more suitable? It would leave the SinchClient as it is and avoid the overhead of creating 2 classes that implement the WebhooksService interface while being completely disjoint.

_Outcome of the discussion_: the "SinchClient" is considered as a "SDK client" to provide access to functionalities offered by the SDK. Regarding the webhooks actions, they are thus also registered on the SinchClient's corresponding domain. The Conversation one is special in the sense it contains API endpoints for webhooks configuration and methods to parse incoming events.
The decision is to keep both functionalities exposed under the "webhooks" domain as it is now. This PR will be updated to not require an instantiation of a HTTPClient when using the `parseEvent` and `validateAuthenticationHeader` methods.

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/WebHooksAdaptorServiceTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+  WebHooksService service;
+
+  @Mock Runnable credentialValidation;
+  @Mock WebHooksApiService apiService;
+  @Mock WebHooksCallbackService callbackService;
+
+  @BeforeEach
+  public void initMocks() {
+    service = spy(new WebHooksAdaptorService(credentialValidation, apiService, callbackService));
```
asein-sinch: The methods  `validateAuthenticationHeader` and `parseEvent` are also exposed on the `WebHooksAdaptorService` class. Do you want to also test them and make sure the `credentialValidation` is not executed?

JPPortier: **Very** **Very** good point 👌 
Done

JPPortier:
Enhanced solution with lazyLoading init to retrieve HttpClient when required: then the webhooks callbacks service (validate, parse) do not require to create one (which is not used for this context)
Will be extended to SMS, Voice, ... in different PRs but for same ticket.
Planned to be released for V1.6.

client/src/main/com/sinch/sdk/domains/conversation/api/templates/adapters/TemplatesService.java
```diff
@@ -40,51 +45,65 @@ public TemplatesService(
...
-    this.context = context;
-    this.httpClient = httpClient;
+      Supplier<HttpClient> httpClientSupplier) {
 
-    OAuthManager authManager =
-        new OAuthManager(credentials, oAuthServer, new HttpMapper(), httpClient);
-    authManagers =
-        Stream.of(new AbstractMap.SimpleEntry<>(SECURITY_SCHEME_KEYWORD_, authManager))
-            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    this.uriUUID = null != credentials ? credentials.getProjectId() : null;
```
asein-sinch, line 50: What's the point of initializing the TemplateService if the credentials are null. It won't be usable anyway.

client/src/main/com/sinch/sdk/domains/conversation/api/v1/adapters/WebHooksAdaptorService.java
```diff
@@ -0,0 +1,90 @@
...
+    if (null == this.apiService) {
+      instanceLazyInit.run();
+      this.apiService =
+          new WebHooksApiService(
+              uriUUID, context, httpClientSupplier.get(), authManagersSupplier.get());
+    }
+    return getApiService();
+  }
+
+  protected WebHooksApiService getApiService() {
```
asein-sinch, line 69: Maybe add a small comment to justify this additional method, introduced for tests.

asein-sinch (approved):

client/src/main/com/sinch/sdk/domains/conversation/api/v1/adapters/WebHooksAdaptorService.java
JPPortier: Will have add such comment everywhere 😢

client/src/main/com/sinch/sdk/domains/conversation/api/templates/adapters/TemplatesService.java
JPPortier: It follow same pattern as for other services (ok there is no special handling required for webhooks templates which are not existing).
And stay open for the future: if in the future there is a V3 for template requiring a dedicated validation: then V3 will validate and no side effect here not V1 and V2.
OCP: open for extension, close for modification

PR #211, opened 2025-02-19
LazyInit for Verification services

JPPortier:


client/src/main/com/sinch/sdk/domains/verification/api/v1/adapters/VerificationService.java
```diff
@@ -93,42 +79,67 @@ private void setApplicationCredentials(ApplicationCredentials credentials) {
...
+      this.statusService =
+          new VerificationStatusService(context, httpClientSupplier.get(), clientAuthManagers);
     }
     return this.statusService;
   }
 
   public WebHooksService webhooks() {
-    checkCredentials();
     if (null == this.webhooks) {
+      instanceLazyInit();
```
asein-sinch, line 109: This is indeed a special one, where the API credentials are the same to validate the webhook signature 👍

asein-sinch (approved):

PR #215, opened 2025-02-21
DEVEXP-744: SMS synch

JPPortier:
- Fix contentType VS content_type JSON field name for MediaMessageBodyDetails at protocol level
- code and status for filtering request are now defined as Array at protocol level
- body requests defined as required at protocol level


openapi-contracts/src/main/com/sinch/sdk/domains/sms/api/v1/adapters/BatchesServiceImpl.java
```diff
@@ -170,6 +170,11 @@ private HttpRequest dryRunRequestBuilder(
       throw new ApiException(
           400, "Missing the required parameter 'this.servicePlanId' when calling dryRun");
     }
+    // verify the required parameter 'sendRequest' is set
+    if (sendRequest == null) {
+      throw new ApiException(
```
Dovchik, line 175: Do you differentiate between actual API Exception with status code returned by the server and API Exception throw only by SDK itself?

Dovchik (approved):
LGTM, besides question which is out of scope of the PR

openapi-contracts/src/main/com/sinch/sdk/domains/sms/api/v1/adapters/BatchesServiceImpl.java
JPPortier: No
Currently this is nearly out of the box what is coming from generator and it mimic what the spec are highlighting: a required parameter is missing so the request is a "bad request"
This could be enhance but we should go to the ADR related to the exceptions

PR #226, opened 2025-04-01
DEVEXP-701: Mailgun e2e templates tests

JPPortier:


client/src/test/java/com/sinch/sdk/e2e/domains/mailgun/v1/TemplatesSteps.java
```diff
@@ -0,0 +1,405 @@
...
+
+    listOnePageResponse = service.list(domainName, request);
+  }
+
+  @When("^I send a request to list all the Templates$")
+  public void listAll() {
+    ListTemplatesQueryParameters request =
+        ListTemplatesQueryParameters.builder().setLimit(2).build();
+
+    listAllResponse = service.list(domainName, request);
```
asein-sinch: I find it a bit confusing to see the same query for:
 - I send a request to list the Templates
 - I send a request to list all the Templates
 - I iterate manually over the Templates pages
 
 and the fact that the logic to fetch additional pages will be performed in the `then` with the `checkExpectedCount` method
(Same comment applies for list versions too)

client/src/test/java/com/sinch/sdk/e2e/domains/mailgun/v1/TemplatesSteps.java
```diff
@@ -0,0 +1,405 @@
...
+  }
+
+  @Then(
+      "the create template with version response contains information about the Template and"
+          + " Version Details")
+  public void createWithVersionResult() {
+
+    // cannot use recursion due to trick to have "version" hidden inside "template"
+    // test field by field
+    compareTemplateField(expectedTemplate, createWithActiveVersionResponse.getLeft());
```
asein-sinch, line 279: Will you be able to use a "normal" recursion check when the fix will be loaded on the backend side?

client/src/test/java/com/sinch/sdk/e2e/domains/mailgun/v1/TemplatesSteps.java
```diff
@@ -0,0 +1,405 @@
...
+    checkExpectedCount(listAllManuallyResponse, expected);
+  }
+
+  @Then("the Templates iteration result contains the data from \"{int}\" pages")
+  public void listAllByPageResult(int expected) {
+
+    int count = listAllManuallyResponse.getContent().isEmpty() ? 0 : 1;
+    while (listAllManuallyResponse.hasNextPage()) {
+      count++;
+      listAllManuallyResponse = listAllManuallyResponse.nextPage();
```
asein-sinch: I'm confused here. When invoking `checkExpectedCount()` on line 307, don't you already iterate through all the pages? Do you need to do it again here?
I understand that you kept the reference to the first pointer in the variable `listAllManuallyResponse`: do you confirm you want to showcase the iteration twice in this scenario?

client/src/test/java/com/sinch/sdk/e2e/domains/mailgun/v1/TemplatesSteps.java
JPPortier: Right
This should be refactored and will be more in respect of "When" and "Then" structure
Done

client/src/test/java/com/sinch/sdk/e2e/domains/mailgun/v1/TemplatesSteps.java
JPPortier: No even when fixed from backend: it would require to have access to protected builder and because of located at different package of these tests; java language do not enable it without, some tricks.... not in scope of e2e

client/src/test/java/com/sinch/sdk/e2e/domains/mailgun/v1/TemplatesSteps.java
JPPortier: This branch of code removed with factoring pushed according to 1st comment

asein-sinch (approved):

PR #227, opened 2025-04-01
Mailgun support (Emails & Templates)

JPPortier:
Sum of all previous PRs related to Mailgun planned for next version: 1.6


asein-sinch (approved):
If I understand well, we find in this PR:
- updates in the HttpClient needed for mailgun pagination
- updates in the configuration for mailgun-specific data (credentials / url / region / storage)
- updates in the core to handle form-data parameters requests (not using jackson)
- the support for mailgun emails + templates subdomains (and samples)

JPPortier:
> If I understand well, we find in this PR:
> 
>     * updates in the HttpClient needed for mailgun pagination
> 
>     * updates in the configuration for mailgun-specific data (credentials / url / region / storage)
> 
>     * updates in the core to handle form-data parameters requests (not using jackson)
> 
>     * the support for mailgun emails + templates subdomains (and samples)

Yes, as big picture, we could also point support for: 
 - RFC822 date format for form parameter serialization but also JSON de-serialization
 - UTC timestamp in second de-serialization
 - URL path fragment encoding enhancement: support properly unicode code points encoding

PR #238, opened 2025-05-14
DEVEXP-889: OAS synchro w19 conversation

JPPortier:
- feat: Support new 'reason.channel_code' field
- fix: Synchronized with fix from specs: TemplateReference with 'required' version


client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/types/template/TemplateReferenceMapper.java
```diff
@@ -0,0 +1,29 @@
...
+
+  static TemplateReference convert(TemplateReferenceInternal _from) {
+    TemplateReferenceInternalImpl from = (TemplateReferenceInternalImpl) _from;
+    TemplateReference.Builder builder = TemplateReference.builder();
+    from.languageCode().ifPresent(builder::setLanguageCode);
+    from.parameters().ifPresent(builder::setParameters);
+    from.templateId().ifPresent(builder::setTemplateId);
+    from.version().ifPresent(builder::setVersion);
+    return builder.build();
+  }
```
asein-sinch, line 28: If it looks like a duck, walks like a duck and quacks like a duck, then it's a duck!
But you need to pay more attention with java, watch inside! 😅

asein-sinch (approved):

client/src/main/com/sinch/sdk/domains/conversation/models/v1/messages/types/template/TemplateReferenceMapper.java
JPPortier: Exact, this should be why this icon was chose for then language: 
image

You need to stay focused, and coffee should help in many situations

PR #247, opened 2025-06-17
DEVEXP-984: Emit warning log onto default regions usage (SMS, Conversation, Mailgun)

JPPortier:


client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -51,10 +52,75 @@ void defaultSmsUrlAvailable() {
   }
 
   @Test
-  void defaultSmsRegion() {
+  void defaultSmsRegionIsUS() {
     Configuration configuration = Configuration.builder().build();
     SinchClient client = new SinchClient(configuration);
-    assertEquals(SMSRegion.US, client.getConfiguration().getSmsContext().get().getSmsRegion());
+    assertEquals(client.getConfiguration().getSmsContext().get().getSmsRegion(), SMSRegion.US);
```
asein-sinch: Nit: although the equality is bijective, the convention is assertEquals(actual, expected). You refactored this line correctly, but the rest of the file is doing the other way around.

asein-sinch: You may also add a comment in this test that it's expected to fail with the 2.0

client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -51,10 +52,75 @@ void defaultSmsUrlAvailable() {
...
+    SinchClient client = new SinchClient(configuration);
+    assertEquals(SMSRegion.EU, client.getConfiguration().getSmsContext().get().getSmsRegion());
+  }
+
+  @Test
+  void defaultConversationUrlAvailable() {
+    Configuration configuration =
+        Configuration.builder().setKeyId("foo").setKeySecret("foo").setProjectId("foo").build();
+    SinchClient client = new SinchClient(configuration);
+    assertNotNull(client.getConfiguration().getConversationContext().get().getUrl());
```
asein-sinch, line 74: As you know which URL it will be, you can check against the value, and add a comment it will be null with the 2.0

client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -51,10 +52,75 @@ void defaultSmsUrlAvailable() {
...
+
+  @Test
+  void defaultConversationTemplateUrlAvailable() {
+    Configuration configuration =
+        Configuration.builder().setConversationRegion(ConversationRegion.EU).build();
+    SinchClient client = new SinchClient(configuration);
+    assertEquals(
+        "https://eu.conversation.api.sinch.com",
+        client.getConfiguration().getConversationContext().get().getUrl());
+  }
```
asein-sinch: This 2 tests are the same. The second one is not in line with its description

client/src/test/java/com/sinch/sdk/SinchClientTest.java
```diff
@@ -150,10 +216,10 @@ void defaultMailgunUrlAvailable() {
   }
 
   @Test
-  void defaultMailgunRegion() {
+  void defaultMailgunRegionIsUS() {
     Configuration configuration = Configuration.builder().build();
     SinchClient client = new SinchClient(configuration);
-    assertNull(client.getConfiguration().getMailgunContext().get().getRegion());
+    assertEquals(client.getConfiguration().getMailgunContext().get().getRegion(), MailgunRegion.US);
   }
```
asein-sinch, line 222: The region used to be null be default. But here you are setting it as "US" by default in order to remove it in the v2? Or did I miss something?

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationServiceTest.java
```diff
@@ -52,4 +65,140 @@ void checkCredentialsTranscoding() {
...
+
+    UnifiedCredentials credentials =
+        UnifiedCredentials.builder()
+            .setKeyId("foo key ")
+            .setKeySecret("foo secret")
+            .setProjectId("foo project")
+            .build();
+    ConversationContext context =
+        ConversationContext.builder()
+            .setUrl("https://sms.foo.url")
```
asein-sinch: Nit: foo value but branded as `sms` in a conversation context

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationServiceTest.java
```diff
@@ -52,4 +65,140 @@ void checkCredentialsTranscoding() {
...
     assertDoesNotThrow(() -> new TemplatesService(null, null, null, null), "Init passed");
   }
+
+  @Test
+  void defaultRegionWarning() {
+
+    LogRecorder logRecorder = new LogRecorder();
+    LogManager.getLogManager().reset();
+    Logger rootLogger = LogManager.getLogManager().getLogger("");
+    rootLogger.addHandler(logRecorder);
```
asein-sinch, line 75: With your test framework, do you have utilities like beforeEach() / AfterEach() to set-up / reset a test context?

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: > Nit: although the equality is bijective, the convention is assertEquals(actual, expected). You refactored this line correctly, but the rest of the file is doing the other way around.

expected/equals: inverted

> You may also add a comment in this test that it's expected to fail with the 2.0

The test name is `defaultSmsRegionIsUS`; I'm pretty sure that when version 2.0 arrives, we'll have the break change in mind when the test fails :-)

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: About checking against the expected value, I'm not in favor of testing against a value coming from configuration file (not hardcoded within sources) even if chance to change the default URL are very very low.

And, like for default region, when the URL value will change for 2.0 it will be part of devs.

Public APIS and logic are tagged as deprecated, not sure we have to spread it across related tests

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: 👌 
Thank you.
Fixed

client/src/test/java/com/sinch/sdk/SinchClientTest.java
JPPortier: 😄 I agree: Yes it could be seems strange

It is due to SinchClient logic changed: before modification, the SinchClient init was not modifying the context setting default region to US( just URL and storages was stored onto sinchclient)
With current modification, same pattern applied to SMS, Conversation and Mailgun: region and flag related to  region are set during sinchclient init... and currently the default region is "US"

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationServiceTest.java
JPPortier: Fixed

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationServiceTest.java
JPPortier: Yes; there is but in this case, this will imply a variable shared across all tests, and in the context (not yet activated) we want to run them in parallel it will cause issue.

Now.... because you raised the point: 
- activate the parallel mode for tests: ✅  .... 😇

client/src/test/java/com/sinch/sdk/SinchClientTest.java
asein-sinch: I don't see any change related to this in your last commits. Did you forget to push something?

JPPortier: It was: related to using "BR" region for template: https://github.com/sinch/sinch-sdk-java/pull/247/commits/65156cb4d20492c8cf0d5f44758afc809617d918#diff-0704faebf7082537987e873145a494b95f7a6d6cf9885cacc1c6f7fa959d8cd3R103

But according to answer made onto https://github.com/sinch/sinch-sdk-java/pull/247#discussion_r2158346538
I just commited same test (just NOT null excepted) for `defaultConversationTemplateUrlAvailable`: https://github.com/sinch/sinch-sdk-java/pull/247/commits/d2ef1f1539ad7abe938bc3909b9b8f73d6013970

asein-sinch (approved):

PR #249, opened 2025-06-19
DEVEXP-740: Switch from 'Basic' auth to 'OAuth' for Numbers

JPPortier:
note: PR opened onto a branch not yet ready for PR but this will reduce conflicts onto cross branches for concurrent devs

### DIFF
--- a/client/src/test/java/com/sinch/sdk/domains/numbers/adapters/CredentialsValidationHelper.java
+++ b/client/src/test/java/com/sinch/sdk/domains/numbers/adapters/CredentialsValidationHelper.java
@@ -5,6 +5,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.sinch.sdk.core.http.HttpClient;
+import com.sinch.sdk.core.models.ServerConfiguration;
 import com.sinch.sdk.models.NumbersContext;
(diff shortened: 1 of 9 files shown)


client/src/test/java/com/sinch/sdk/domains/numbers/adapters/CredentialsValidationHelper.java
```diff
@@ -27,10 +28,13 @@ static void doNotAcceptNullKey(
     UnifiedCredentials credentials =
         UnifiedCredentials.builder().setKeyId(null).setKeySecret("foo").setProjectId("foo").build();
     NumbersContext context = NumbersContext.builder().setNumbersUrl("foo url").build();
+    ServerConfiguration oAuthServer = new ServerConfiguration("https://oauth.foo.url");
```
asein-sinch, line 31: In this test file, you are checking the credentials definition, not the auth server. Instead of repeating the same definition in each test, you may define it once and share it with all the tests.

asein-sinch (approved):

client/src/test/java/com/sinch/sdk/domains/numbers/adapters/CredentialsValidationHelper.java
JPPortier: Good point.
Done

PR #251, opened 2025-07-02
build: Added '@Override' annotation for generated services

JPPortier:


asein-sinch (approved):
Shipping 2 updates in the same PR to give the impression the numbers of files is the same, smart! 👀

PR #254, opened 2025-07-15
DEVEXP-547: e2e Verification

JPPortier:


client/src/test/java/com/sinch/sdk/e2e/domains/verification/v0/ReportSteps.java
```diff
@@ -0,0 +1,158 @@
...
+  VerificationReport reportSmsByIdentityResponse;
+  VerificationReport reportPhoneCallByIdResponse;
+  VerificationReport reportPhoneCallByIdentityResponse;
+  VerificationReport reportFlashCallByIdResponse;
+  VerificationReport reportFlashCallByIdentityResponse;
+
+  @Given("^the Verification service \"Report\" is available$")
+  public void serviceAvailable() {
+    service = Config.getSinchClient().verification().verifications();
+    ;
```
matsk-sinch: remove ;

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v1/ReportSteps.java
```diff
@@ -0,0 +1,161 @@
...
+
+  @Then(
+      "the response by \"identity\" contains the details of a failed Flash Call verification"
+          + " report")
+  public void reportFlashCallFailedResult() {
+
+    TestHelpers.recursiveEquals(
+        reportFlashCallByIdentityResponse,
+        VerificationReportResponseFlashCall.builder()
+            .setId("1ce0ffee-c0de-5eed-d22d-f00dfeed1337")
```
matsk-sinch, line 155: I'm wondering if you should set these variables as a constant?

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v1/ReportSteps.java
```diff
@@ -0,0 +1,161 @@
...
+            .setStatus(VerificationStatus.FAIL)
+            .setCallComplete(true)
```
matsk-sinch, line 157: should callComplete be true on failure?

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v0/StartSteps.java
```diff
@@ -0,0 +1,173 @@
...
+
+  VerificationsService service;
+  StartVerificationResponse startVerificationBySmsResponse;
+  StartVerificationResponse startVerificationByPhoneCallResponse;
+  StartVerificationResponse startVerificationByFlashCallResponse;
+  ApiException startSeamlessException;
+
+  @Given("^the Verification service \"Start\" is available$")
+  public void serviceAvailable() {
+    service = Config.getSinchClient().verification().verifications();
```
matsk-sinch, line 38: Should you assert service != null?

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v0/ReportSteps.java
JPPortier: Thank you
Removed

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v1/ReportSteps.java
JPPortier: By regarding sources, this id value is used twice: one for error one for success.
Then we could say: it should be more in charge of mockserver to return here 2 different values instead of a single common one 😃 
Because of other report IDs are differents, let says here we do not need a global const.

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v1/ReportSteps.java
JPPortier: mockserver's payloads are coming from real production servers (not manually forged); then "yes"
We can imagine in some context this could be set to false then the verification failed because call was not completed and having dedicated status/reason fields providing more info about the cause of failure

client/src/test/java/com/sinch/sdk/e2e/domains/verification/v0/StartSteps.java
JPPortier: I do not think: if null all tests will fails later; it is "just" a matter of init

matsk-sinch (approved):

PR #270, opened 2025-08-28
DEVEXP-1093: W35 OAS synch conversation

JPPortier:


openapi-contracts/src/test/resources/domains/conversation/v1/credentials/ConversationChannelSinchChatRequestDto.json
```diff
@@ -0,0 +1,5 @@
+{
+  "channel": "SINCH_CHAT",
+  "callback_secret": "callback secret",
+  "credential_ordinal_number": 0
+}
```
asein-sinch, line 5: I know this test is purely for technical validation, but I don't think the API would accept such payload, as there are no credentials associated with this channel. My recommendation would be to revisit this code once the documentation team will have clarified how to handle SINCH_CHAT within this API

asein-sinch (approved):

PR #283, opened 2025-11-04
W43 catch up mailgun

JPPortier:
'resend' end point support


client/src/main/com/sinch/sdk/domains/mailgun/api/v1/adapters/ArrayToCommaSeparatedFormSerializer.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Map;
+
+public class ArrayToCommaSeparatedFormSerializer extends FormSerializer<List<String>> {
+
+  @Override
+  public void serialize(List<String> in, String fieldName, Map<String, Object> out) {
+    if (null != in) {
+      out.put(fieldName, String.join(",", in));
+    }
+  }
```
asein-sinch, line 14: I just want to be sure about this method: it will concatenate all the list entries and put a comma between them. You have covered in your tests the fact a list entry can contain a comma-separated list of values. 
That's a great edge case to have considered! Would you like to split this kind of entries and trim the data before doing the concatenation?

client/src/test/java/com/sinch/sdk/domains/mailgun/api/v1/adapters/ArrayToCommaSeparatedFormSerializerTest.java
asein-sinch, line 1: What about the entry `Arrays.asList("value1", "value2,,,,,")`
Should it be serialized into `value1,value2,,,,,` or `value1,value2`

sample-app/src/main/java/com/sinch/sample/mailgun/messages/ReSend.java
```diff
@@ -0,0 +1,139 @@
...
+      LOGGER.severe(e.getMessage());
+      e.printStackTrace();
+    }
+  }
+
+  public void run() {
+
+    EmailsService service = client.mailgun().v1().emails();
+
+    LOGGER.info("ReSend email with Mailgun to: " + mailgunTo);
```
asein-sinch, line 30: Where is defined this variable?

sample-app/src/main/java/com/sinch/sample/mailgun/messages/ReSend.java
```diff
@@ -0,0 +1,139 @@
...
+    String storageKey = "storage key value";
+
+    ResendRequest parameters = ResendRequest.builder().setTo(Arrays.asList(mailgunTo)).build();
+
+    SendEmailResponse response = service.resend(mailgunDomain, storageKey, parameters);
+
+    LOGGER.info("Response: " + response);
+  }
+
+  SendEmailRequest createTextEmail() {
```
asein-sinch: Why do you need this method in the ReSend class ?

client/src/main/com/sinch/sdk/domains/mailgun/api/v1/adapters/ArrayToCommaSeparatedFormSerializer.java
JPPortier: If there is such constraint from backend to respect this: yes we could have to implement this as "helper.

But there is no such requirements from backend; then no justification to add dedicated business logic in front of request

client/src/test/java/com/sinch/sdk/domains/mailgun/api/v1/adapters/ArrayToCommaSeparatedFormSerializerTest.java
JPPortier: No dedicated (nor required by mailgun API) business logic here; then `value1,value2,,,,,` is expected

Added dedicated tests to "document" it

sample-app/src/main/java/com/sinch/sample/mailgun/messages/ReSend.java
JPPortier: Read from configuration file: https://github.com/sinch/sinch-sdk-java/blob/feat/V2.0-next/sample-app/src/main/resources/config.properties#L18

And available from test because if inherit of `BaseApplication`: https://github.com/sinch/sinch-sdk-java/blob/feat/V2.0-next/sample-app/src/main/java/com/sinch/sample/BaseApplication.java#L64

sample-app/src/main/java/com/sinch/sample/mailgun/messages/ReSend.java
JPPortier: 🤦 
Removed.
Thank you

asein-sinch (approved):

PR #301, opened 2026-02-02
Read logging configuration file from snippets

JPPortier:


matsk-sinch (approved):

examples/snippets/src/main/java/utils/Settings.java
```diff
@@ -70,4 +82,17 @@ public static Optional<String> getConversationRegion() {
   public static Optional<String> getPhoneNumber() {
     return get("SINCH_PHONE_NUMBER");
   }
+
+  public static void initializeLogger() {
```
asein-sinch: This can be made private

examples/snippets/src/main/java/utils/Settings.java
```diff
@@ -70,4 +82,17 @@ public static Optional<String> getConversationRegion() {
...
+    try (InputStream logConfigInputStream =
+        Settings.class.getClassLoader().getResourceAsStream(LOGGING_CONFIG_FILE)) {
+      if (logConfigInputStream != null) {
+        LogManager.getLogManager().readConfiguration(logConfigInputStream);
+      }
+    } catch (IOException ioe) {
+      LOGGER.config(
+          String.format(
+              "Error reading '%s' configuration file: %s", LOGGING_CONFIG_FILE, ioe.getMessage()));
```
asein-sinch: You are using the logger to log an error about the logger that may have not been initialized. Is it ok to do that, or the error message will be invisible?

examples/snippets/src/main/java/utils/Settings.java
```diff
@@ -1,20 +1,32 @@
...
+
   private static final Properties properties;
 
   static {
+    initializeLogger();
     properties = new Properties();
     try {
       // load a properties file from class path, inside static method
-      properties.load(Settings.class.getClassLoader().getResourceAsStream("config.properties"));
+      properties.load(Settings.class.getClassLoader().getResourceAsStream(CONFIG_FILE));
```
asein-sinch: Nit: instead of throwing a NPE when the file is not found, you may check if the inputStream is not null before loading the properties and log that the config file is not found if null

examples/snippets/src/main/java/utils/Settings.java
JPPortier: Done

examples/snippets/src/main/java/utils/Settings.java
JPPortier: Changed severity to "warning"; log is available

examples/snippets/src/main/java/utils/Settings.java
JPPortier: done (and applied to to log configuration reading)

asein-sinch (approved):

examples/snippets/src/main/java/utils/Settings.java
```diff
@@ -70,4 +89,19 @@ public static Optional<String> getConversationRegion() {
...
     return get("SINCH_PHONE_NUMBER");
   }
+
+  private static void initializeLogger() {
+    try (InputStream logConfigInputStream =
+        Settings.class.getClassLoader().getResourceAsStream(LOGGING_CONFIG_FILE)) {
+      if (logConfigInputStream != null) {
+        LogManager.getLogManager().readConfiguration(logConfigInputStream);
+      } else {
+        LOGGER.warning(String.format("Error reading '%s' configuration file", LOGGING_CONFIG_FILE));
```
rpredescu-sinch, line 99: Hi, @JPPortier!
Do you think if better to have same severity log of as above (`LOGGER.severe`) when configuration file can not be read?
Best!

JPPortier: Don't think.

Logs are not critical/required when running snippets.

But without proper credentials set it could be (even if snippet file can override value)

rpredescu-sinch (approved):

PR #310, opened 2026-02-25
DEVEXP-1278: Using supplier in place of services DI for pagination

JPPortier:


openapi-contracts/src/main/com/sinch/sdk/domains/sms/models/v1/batches/response/ListBatchesResponse.java
```diff
@@ -37,24 +36,18 @@ public boolean hasNextPage() {
       return false;
     }
     if (null == nextPage) {
-      ListBatchesQueryParameters.Builder newParameters =
-          ListBatchesQueryParameters.builder(page.getParameters());
-      newParameters.setPage(page.getNextPageToken());
-      nextPage = service.list(newParameters.build());
+      nextPage = supplier.get();
```
rpredescu-sinch, line 39: Hi @JPPortier!

If the next-page token is present, the supplier is called eagerly and the result is cached in nextPage:

```
if (null == nextPage) {
    nextPage = supplier.get();  // HTTP call #1 — result cached
}
```

But `nextPage()` then clears the cache and calls the supplier again:

```
nextPage = null;
return supplier.get();  // HTTP call #2
```

This means every hasNextPage() + nextPage() pair makes two HTTP requests for the same page.

We can test this by adding the following in the list test in `BatchesServiceTest.java`:

```
@Test
void listDoesNotFetchNextPageTwiceWhenHasNextPageCalledBeforeNextPage()
    throws ApiException {

  // ... (set up page0 and page1 HttpRequest/HttpResponse mocks as in the existing list() test)

  ListBatchesResponse page0 = service.list(initialRequest);

  page0.hasNextPage(); // triggers 1st fetch of page 1 and caches it
  page0.nextPage();    // should return cached result — but currently fetches again

  // Fails: Wanted 1 time but was 2 times
  verify(httpClient, times(1)).invokeAPI(
      eq(serverConfiguration), eq(authManagers),
      argThat(new HttpRequestMatcher(httpRequest1)));
}
```

The current implementation of `hasNextPage()` is performing a HTTP request callers might not expect from a boolean check.

Maybe the right fix is to follow the conversation `EventsListResponse.java` implementation which is very simple and relies on the page to tell us if there is any next page.

Affected classes are: `ListBatchesResponse`, `ListGroupsResponse`, `ListDeliveryReportsResponse` and `ListInboundsResponse`.

What do you think?

Best!

core/src/test/java/com/sinch/sdk/core/TestHelpers.java
```diff
@@ -24,4 +43,16 @@ public static void checkIteratorItems(Iterator<?> iterator, int expectedCount) {
     iterator.forEachRemaining(i -> count.incrementAndGet());
     Assertions.assertThat(count.get()).isEqualTo(expectedCount);
   }
+
+  static boolean isFunctionalInterface(Class<?> iface) {
```
rpredescu-sinch: Hi @JPPortier!

The `isFunctionalInterface()` method detects functional interfaces by counting abstract methods, but this also matches single-abstract-method interface like `Comparable` or `Iterable`, not just lambda fields. A field of type `Iterable` will be silently excluded from recursive comparison.

An `Iterable` field in a class can hold a concrete collection which should be compared?

Since @FunctionalInterface can only be applied to interfaces, the `isInterface()` guard and the abstract method counting logic can both go away.

```
private static boolean isFunctionalInterface(Class<?> type) {
    return type.isAnnotationPresent(FunctionalInterface.class);
}
```

Or we can use it inline?

```
Arrays.stream(actual.getClass().getDeclaredFields())
    .filter(f -> f.getType().isAnnotationPresent(FunctionalInterface.class))
    .map(Field::getName)
    .toArray(String[]::new)
```

What do you think?

openapi-contracts/src/main/com/sinch/sdk/domains/sms/models/v1/batches/response/ListBatchesResponse.java
```diff
@@ -12,22 +12,21 @@
...
 import java.util.NoSuchElementException;
+import java.util.function.Supplier;
 
 /** Auto paginated response for list of BatchResponse */
 public class ListBatchesResponse extends ListResponse<BatchResponse> {
 
-  private final Page<ListBatchesQueryParameters, BatchResponse, Integer> page;
-  private final BatchesService service;
+  private final Page<BatchResponse, Integer> page;
+  final Supplier<ListBatchesResponse> supplier;
```
rpredescu-sinch, line 24: Hi @JPPortier!

Should this be `private final` rather than `package-private`? The missing access modifier means it's visible to the entire package.

Best!

openapi-contracts/src/main/com/sinch/sdk/domains/numbers/models/v1/response/AvailableNumbersListResponse.java
```diff
@@ -10,45 +10,28 @@
...
 import java.util.Collection;
 import java.util.Collections;
-import java.util.NoSuchElementException;
 
 /** Auto paginated response for list of AvailableNumber */
 public class AvailableNumbersListResponse extends ListResponse<AvailableNumber> {
 
-  private final Page<AvailableNumbersListQueryParameters, AvailableNumber, HttpRequest> page;
-  private final AvailableNumberServiceImpl service;
+  private final Page<AvailableNumber, Object> page;
```
rpredescu-sinch, line 21: Hi, @JPPortier!

`Object` might be too generic here? 
Could `Void` be used instead? Unlike `Object`, it explicitly signals that the token will always be `null` and never used.

What do you think?

If yes, these classes can have the same update: `AppsListResponse, WebhooksListResponse, TemplatesV2ListResponse, AvailableRegionsListResponse, OwnedNumbersListResponse`.

Best!

MIGRATION-GUIDE.md
```diff
@@ -373,21 +373,6 @@ Use the new versioned API under `voice().v1()` to get access to [VoiceService](h
 
 Use the new versioned API under `verification().v1()` to get access to [VerificationService](https://www.javadoc.io/doc/com.sinch.sdk/sinch-sdk-java/2.0.0/com/sinch/sdk/domains/verification/api/v1/VerificationService.html): [`sinchClient.verification().v1()`](https://www.javadoc.io/doc/com.sinch.sdk/sinch-sdk-java/2.0.0/com/sinch/sdk/domains/verification/VerificationService.html#v1())
 
-| Old interface                         | New Equivalent in `verification().v1()`                                                                                                                               |
```
rpredescu-sinch, line 376: Hi @JPPortier!

Just checking if #### Replacement APIs section deleted on purpose. Not a valid migration any longer?

Thank you!

openapi-contracts/src/main/com/sinch/sdk/domains/sms/models/v1/batches/response/ListBatchesResponse.java
JPPortier: Very good catch!

> Maybe the right fix is to follow the conversation EventsListResponse.java implementation which is very simple and relies on the page to tell us if there is any next page.

This is not possible the same way: with token pagination like with Conversation this can be done but here it is  SMS and page offset navigation

The issue here is about page offset and the response from backend alone cannot be used to detect if there is another page: this require to check response against query parameters **and** response.

I will create a separate PR for this point (and will work on it just after this PR)

openapi-contracts/src/main/com/sinch/sdk/domains/sms/models/v1/batches/response/ListBatchesResponse.java
JPPortier: Right (will be part of following PR to avoid conflict with current stacked and because cross domain one)

openapi-contracts/src/main/com/sinch/sdk/domains/numbers/models/v1/response/AvailableNumbersListResponse.java
JPPortier: Indeed this is a good use case Void 👌 

Will be part of following PR

MIGRATION-GUIDE.md
JPPortier: Yes.

See just below the section was present twice

core/src/test/java/com/sinch/sdk/core/TestHelpers.java
JPPortier: Why doing complicated when an simple implementation is possible !

Thank you. 

Done

rpredescu-sinch (approved):
Good to go!

PR #329, opened 2026-03-20
DEVEXP-987: remove support of default region

JPPortier:


client/src/main/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationService.java
```diff
@@ -249,20 +249,13 @@ private void instanceLazyInit() {
             credentials.getKeySecret(), "Conversation service requires 'keySecret' to be defined");
         StringUtil.requireNonEmpty(
             credentials.getProjectId(), "Conversation service requires 'projectId' to be defined");
+        Objects.requireNonNull(
+            context.getRegion(), "Conversation service requires 'region' to be defined");
+        StringUtil.requireNonEmpty(
```
rpredescu-sinch, line 254: Hi @JPPortier!

Another idea would be to add `StringUtil.requireNonEmpty` in `ConversationRegion.from` and remove it from here.

```
public static ConversationRegion from(String value) {
    StringUtil.requireNonEmpty(value, "ConversationRegion value cannot be empty");
    return ENUM_SUPPORT.from(value);
}
```

This way we do not have to add the string validation anywhere else. Same for `SMSService`.

Best!

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationServiceTest.java
```diff
@@ -68,143 +55,13 @@ void checkCredentialsTranscoding() {
...
-
-    Optional<LogRecord> record =
-        logRecorder.records.stream()
-            .filter(
-                r -> r.getMessage().contains("Using default region for Conversation Template 'us"))
-            .findFirst();
-
-    assertFalse(record.isPresent());
-    LogManager.getLogManager().reset();
+  void passInitWithoutSettings() {
```
rpredescu-sinch, line 63: Hi, @JPPortier!

Could be better to have a dedicated TemplatesServiceTest.java with this test (and any future template-specific tests)? Nice to have. 

Best!

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/CredentialsValidationHelper.java
rpredescu-sinch, line 1: Hi @JPPortier !
This file name do not have the `Test` suffix as the others do.
Best!

client/src/main/com/sinch/sdk/domains/conversation/templates/api/adapters/TemplatesService.java
```diff
@@ -82,22 +82,12 @@ private void instanceLazyInit() {
             credentials.getKeySecret(), "Templates service requires 'keySecret' to be defined");
         StringUtil.requireNonEmpty(
             credentials.getProjectId(), "Templates service requires 'projectId' to be defined");
-
+        Objects.requireNonNull(
```
rpredescu-sinch, line 85: Hi, @JPPortier!
Only `requireNonNull(region)` was added. Do we also need `requireNonEmpty(region.value())`?
Best!

client/src/test/java/com/sinch/sdk/domains/conversation/api/templates/adapters/CredentialsValidationHelper.java
```diff
@@ -147,6 +151,26 @@ private static void doNotAcceptNullURLTemplates(
             .contains("Templates service requires 'templateManagementUrl' to be defined"));
   }
 
+  private static void doNotAcceptNullRegionTemplates(
```
rpredescu-sinch, line 155: Hi @JPPortier !
This test only covers the null region case. Should we also have a test for empty region `ConversationRegion.from("")`?
Best!

client/src/main/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationService.java
JPPortier: Good point.
The general pattern of dynamic enums here across all of the SDK is to accept any new value without any dedicated business logic (same pattern applied for values coming from backend for example onto responses)

Then by accepting an empty value we can have dedicated business logic when it is required for usage (here during init of service which is checking it own requirements).

If another location (e.g. SDK user sources level) do not have same requirements, an empty value can be accepted without issue.

So in my opinion, it is better to let business logic validation where it have more sense: on use

client/src/main/com/sinch/sdk/domains/conversation/templates/api/adapters/TemplatesService.java
JPPortier: 👌 
Fixed

client/src/test/java/com/sinch/sdk/domains/conversation/api/templates/adapters/CredentialsValidationHelper.java
JPPortier: Good catch 👌 
added

client/src/test/java/com/sinch/sdk/domains/conversation/api/v1/adapters/ConversationServiceTest.java
JPPortier: 🤦 
cut/paste issue.
Test is really for ConversationService (related Templates one  already exist onto TemplateServiceTest)

Updated

client/src/test/java/com/sinch/sdk/domains/sms/api/v1/adapters/CredentialsValidationHelper.java
JPPortier: Yes.

Like other CredentialsValidationHelpers files (SMS, Conversation, ConversationTemplate): it do not contains any test declaration, just helpers called by them

asein-sinch (approved):

PR #333, opened 2026-03-26
Jpportier/pre v2.0 cleanup

JPPortier:


client/src/main/com/sinch/sdk/domains/verification/api/v1/adapters/VerificationService.java
```diff
@@ -71,63 +71,81 @@ private void createAuthManagers(ApplicationCredentials credentials) {
     // hidden behind the "Basic" keyword
     // We need both auth manager to handle Sinch Events because of customer will choose from his
     // dashboard which scheme to be used
-    clientAuthManagers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-    clientAuthManagers.put(
+    Map<String, AuthManager> localClientAuthManagers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    localClientAuthManagers.put(
         BASIC_SECURITY_SCHEME_KEYWORD_VERIFICATION,
         useApplicationAuth ? applicationAuthManager : basicAuthManager);
+    clientAuthManagers = localClientAuthManagers;
```
asein-sinch, line 78: Why do you need to introduce a local variable?

asein-sinch (approved):

client/src/main/com/sinch/sdk/domains/verification/api/v1/adapters/VerificationService.java
JPPortier: Big picture: Should not be required because of private function called from a synch block.

But, here a matter a best practice, and protection if any cut/paste in the future.

Because of map is declared volatile, as soon as the reference of the map is changed, other threads will see the new reference... but we still (`put`) modify map content after init to fill it.

So to prevent other threads consuming a not yet fully inited map, the local variable delayed the other threads propagation mechanism about new reference of map

